PR #1, opened 2019-12-15
Remove PoW consensus

sgkim126:
It fixes https://github.com/CodeChain-io/foundry/issues/4


majecty:
This is so satisfying.

core/src/consensus/mod.rs
```diff
@@ -265,7 +239,6 @@ pub trait ConsensusEngine: Sync + Send {
         header.hash()
     }
 
-    /// In PoW consensus, the higher scored block becomes the best block.
```
majecty: We may remove the "score" in the header. When using the Tendermint consensus algorithm, we can't define a valid "score" value. A finalized block must have a higher score than a proposal block. Since the content of a block is not changed whether it is finalized or not. The score always has a weird value.

core/src/miner/miner.rs
```diff
@@ -412,50 +395,29 @@ impl Miner {
 
     /// Prepares work which has to be done to seal.
     fn prepare_work(&self, block: ClosedBlock, original_work_hash: Option<H256>) {
-        let (work, is_new) = {
-            let mut sealing_work = self.sealing_work.lock();
-            let last_work_hash = sealing_work.queue.peek_last_ref().map(|pb| pb.block().header().hash());
+        let mut sealing_work = self.sealing_work.lock();
```
majecty: @sgkim126 What do you think about removing SealingWork and SealingQueue. IMO, they were used by PoW consensus to try to make multiple blocks' seal at the same time. When we are using the Tendermint Consensus, we only create a seal at a time.
Also in another point of view, the sealwing_work.queue was needed because sealing in the PoW consensus is an asynchronous job. Creating a seal in the Tendermint consensus is a blocking job. We do not need to use the queue.

sgkim126: Looks good. I'll try.

core/src/consensus/mod.rs
sgkim126: I made an issue for it.

sgkim126:
Travis failed because it pulls the pr for [19](https://travis-ci.com/CodeChain-io/foundry/jobs/269594987#L62). I don't know why but I think it's the fault of Travis.

Issue #8, opened 2019-12-19
Simplify mempool


sgkim126:
Remove timelock related features from the mempool.

kseo:
@majecty can we close this issue after PR #120 is landed? or do we have more things to do?

majecty:
I don't know the Memory pool code well. @sgkim126 Can we close this issue after #120 is landed?

sgkim126:
@majecty Yes, I think so.

PR #12, opened 2019-12-19
Remove TxTimlock from the mempool

Remagpie:
Resolves https://github.com/CodeChain-io/foundry/issues/8
I think there might be some parts that can be further simplified, but I'm not sure about them for now.


core/src/miner/mem_pool.rs
```diff
@@ -600,61 +585,13 @@ impl MemPool {
...
     /// Checks the timelock of transactions starting from `start_seq`.
     /// Returns the next seq of the last transaction which can be in the current queue
-    fn check_transactions(
-        &self,
-        public: Public,
-        mut start_seq: u64,
-        current_block_number: PoolingInstant,
-        current_timestamp: u64,
-    ) -> u64 {
+    fn check_transactions(&self, public: Public, start_seq: u64) -> u64 {
```
sgkim126: Why does this function still remain?
It seems that this function was introduced because of a timelock.

core/src/miner/mem_pool.rs
```diff
@@ -269,7 +268,10 @@ impl MemPool {
 
             let id = self.next_transaction_id;
             self.next_transaction_id += 1;
-            let item = MemPoolItem::new(tx, origin, inserted_block_number, inserted_timestamp, id, timelock);
+            let item = MemPoolItem::new(tx, origin, inserted_block_number, inserted_timestamp, id, TxTimelock {
```
sgkim126: I think rebasing this PR on https://github.com/CodeChain-io/foundry/pull/11 will make this commit clear.

core/src/miner/mem_pool.rs
Remagpie: I think this function can be removed, but I couldn't since I didn't fully understand the mempool implementation yet.
It's returning some value derived from `start_seq`, and I'm not sure what it means.

core/src/miner/mem_pool.rs
Remagpie: I'll try it when #11 is merged.

Remagpie:
Rebased to the latest master.

majecty:
I created a new PR #120 with the same commits as this PR. I rebased these commits in the master branch.

PR #18, opened 2019-12-20
Read term_common_params from the state

Remagpie:
Copied from https://github.com/CodeChain-io/codechain/pull/1895
Some parts are revised to match the foundry's situation.

### DIFF
--- a/state/src/impls/top_level.rs
+++ b/state/src/impls/top_level.rs
@@ -999,6 +999,12 @@ impl TopState for TopLevelState {
         metadata.increase_seq();
         Ok(())
     }
+
+    fn update_term_params(&mut self) -> StateResult<()> {
+        let mut metadata = self.get_metadata_mut()?;
+        metadata.update_term_params();
+        Ok(())
+    }
 }
 
 fn is_active_account(state: &dyn TopStateView, address: &Address) -> TrieResult<bool> {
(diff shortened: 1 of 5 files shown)


state/src/item/metadata.rs
```diff
@@ -34,6 +34,7 @@ pub struct Metadata {
     term: TermMetadata,
     seq: u64,
     params: Option<CommonParams>,
+    term_params: Option<CommonParams>,
```
sgkim126: What's the difference between `params` and `term_params`?

Remagpie: `term_params` is the common parameter at the start of the term.
Some parts in our implementation (e.g. tendermint) depend on this value.

state/src/impls/top_level.rs
```diff
@@ -999,6 +999,12 @@ impl TopState for TopLevelState {
         metadata.increase_seq();
         Ok(())
     }
+
+    fn snapshot_term_params(&mut self) -> StateResult<()> {
```
sgkim126: I don't think snapshot is appropriate for this function. How about renaming it to `update_term_params`?

Remagpie: Agreed. It was not my decision to name this function like this, but I think your suggestion is better.

PR #22, opened 2019-12-23
Refactoring patches from CodeChain/snapshot

Remagpie:
Ported from https://github.com/CodeChain-io/codechain/pull/1921

### DIFF
--- a/test/src/helper/spawn.ts
+++ b/test/src/helper/spawn.ts
@@ -861,7 +861,7 @@ export default class CodeChain {
         while (true) {
             const termMetadata = await stake.getTermMetadata(this.sdk);
             if (termMetadata && termMetadata.currentTermId >= target) {
-                break;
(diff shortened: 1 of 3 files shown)


test/src/helper/spawn.ts
```diff
@@ -859,14 +859,16 @@ export default class CodeChain {
     public async waitForTermChange(target: number, timeout?: number) {
         const start = Date.now();
         while (true) {
-            const termMetadata = await stake.getTermMetadata(this.sdk);
+            const termMetadata = (await stake.getTermMetadata(this.sdk))!;
             if (termMetadata && termMetadata.currentTermId >= target) {
```
sgkim126:
```ts
if (termMetadata.currentTermId >= target) {
```

Remagpie: Fixed.

PR #27, opened 2019-12-24
Update tendermint implementation to be compatible with snapshot sync

Remagpie:
Collected from https://github.com/CodeChain-io/codechain/pull/1900 and https://github.com/CodeChain-io/codechain/pull/1922

### DIFF
--- a/core/src/consensus/stake/action_data.rs
+++ b/core/src/consensus/stake/action_data.rs
@@ -410,28 +410,28 @@ impl IntoIterator for Validators {
 
 #[derive(Default, Debug, PartialEq)]
 pub struct IntermediateRewards {
-    previous: BTreeMap<Address, u64>,
     current: BTreeMap<Address, u64>,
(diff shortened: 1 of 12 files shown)


core/src/consensus/stake/action_data.rs
```diff
@@ -410,28 +410,28 @@ impl IntoIterator for Validators {
 
 #[derive(Default, Debug, PartialEq)]
 pub struct IntermediateRewards {
-    previous: BTreeMap<Address, u64>,
-    current: BTreeMap<Address, u64>,
+    pub(super) current: BTreeMap<Address, u64>,
+    pub(super) calculated: BTreeMap<Address, u64>,
```
majecty: Could you explain to me why you used `pub(super)`?

Remagpie: It came from the original PR in CodeChain since `IntermediateRewards` was wrapped inside a module.
We don't need it anymore. I'll remove this.

PR #19, opened 2019-12-20
Create snapshots

Remagpie:
Collected from https://github.com/CodeChain-io/codechain/pull/1869 and https://github.com/CodeChain-io/codechain/pull/1875

PR #29, opened 2019-12-26
Import the snapshot header

Remagpie:
Depends: https://github.com/CodeChain-io/foundry/pull/19
Collected from https://github.com/CodeChain-io/codechain/pull/1855 and https://github.com/CodeChain-io/codechain/pull/1862

### DIFF
--- a/foundry/foundry.yml
+++ b/foundry/foundry.yml
@@ -252,6 +252,18 @@ args:
         takes_value: true
         conflicts_with:
             - no-discovery
+    - snapshot-hash:
+        long: snapshot-hash
+        value_name: HASH
+        help: The block hash of the snapshot target block.
+        requires: snapshot-number
+        takes_value: true
+    - snapshot-number:
+        long: snapshot-number
+        value_name: NUM
+        help: The block number of the snapshot target block.
+        requires: snapshot-hash
+        takes_value: true
     - no-snapshot:
         long: no-snapshot
         help: Disable snapshots
--- a/sync/src/block/extension.rs
+++ b/sync/src/block/extension.rs
@@ -129,10 +129,12 @@ impl Extension {
         let mut header = client.best_header();
         let mut hollow_headers = vec![header.decode()];
         while client.block_body(&BlockId::Hash(header.hash())).is_none() {
-            header = client
-                .block_header(&BlockId::Hash(header.parent_hash()))
-                .expect("Every imported header must have parent");
-            hollow_headers.push(header.decode());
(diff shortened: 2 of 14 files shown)


sync/src/block/extension.rs
```diff
@@ -103,6 +118,45 @@ impl Extension {
         }
     }
 
+    fn initial_state(client: Arc<Client>, snapshot_target: Option<(H256, u64)>) -> State {
```
sgkim126: How about making it a constructor of `State`?

Remagpie: Actually there's a patch that moves this function to `State::initial` method. I'll move that patch to here.

util/merkle/src/triedbmut.rs
```diff
@@ -417,6 +417,10 @@ impl<'a> Trie for TrieDBMut<'a> {
 
         t.get(key)
     }
+
+    fn is_complete(&self) -> bool {
+        TrieDB::try_new(self.db, self.root).map(|t| t.is_complete()).unwrap_or(false)
```
sgkim126:
```rust
TrieDB::try_new(self.db, self.root).map_or(false, |t| t.is_complete())
```

Remagpie: [`Result`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) doesn't have `map_or` function.

sgkim126: https://doc.rust-lang.org/nightly/std/result/enum.Result.html#method.map_or Sorry. It is nightly yet.

foundry/foundry.yml
```diff
@@ -252,6 +252,16 @@ args:
         takes_value: true
         conflicts_with:
             - no-discovery
+    - snapshot-hash:
```
sgkim126: Would you describe what this option does?

foundry/foundry.yml
```diff
@@ -252,6 +252,16 @@ args:
...
+        long: snapshot-hash
+        value_name: HASH
+        requires: snapshot-number
+        takes_value: true
+    - snapshot-number:
```
sgkim126: ditto.

foundry/foundry.yml
Remagpie: Added help message to this option.

foundry/foundry.yml
Remagpie: Added a help message for this option.

PR #34, opened 2019-12-27
Follow-up patches for the tendermint consensus with snapshot sync

Remagpie:
https://github.com/CodeChain-io/foundry/pull/27 introduced a behavior change in the consensus implementation, and some additional problems were found while testing snapshot synchronization.
Detailed explanation is recorded in the commit messages.

### DIFF
--- a/core/src/client/client.rs
+++ b/core/src/client/client.rs
@@ -481,6 +481,10 @@ impl StateInfo for Client {
 }
 
 impl EngineInfo for Client {
+    fn network_id(&self) -> NetworkId {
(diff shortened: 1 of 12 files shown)


Remagpie:
I think I'm a bit confused about the relation between signers and the state item `CurrentValidators`.
It's really confusing that I thought :sob:

Remagpie:
It seems there some problems with this PR. It's not working with the snapshot synchronization.
I'll request the review again when I'm confident with the patch.
Sorry for the inconvenience. @sgkim126 @majecty

PR #36, opened 2019-12-27
Remove asset RPCs

majecty:
Foundry has a module system to make a user be able to create its logic using it. If the user needs a custom asset, they will implement it using the module. So Foundry does not need the Asset system.

I'll create a few more PRs to remove all the Asset code.

~This PR depends on #26~


rpc/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2019 Kodebox, Inc.
+// Copyright 2018 Kodebox, Inc.
```
sgkim126: Would you check it?

sgkim126:
IMO we should remove some e2e tests also. Are there tests that use the removed RPCs?

rpc/src/lib.rs
majecty: Sure. I fixed it.

rpc/src/v1/impls/chain.rs
```diff
@@ -14,36 +14,33 @@
...
-    AccountData, AssetClient, BlockId, EngineInfo, ExecuteClient, MiningBlockChainClient, Shard, TermInfo, TextClient,
-};
-use ccrypto::Blake;
+use ccore::{AccountData, BlockId, EngineInfo, ExecuteClient, MiningBlockChainClient, Shard, TermInfo, TextClient};
 use cjson::scheme::Params;
 use cjson::uint::Uint;
 use ckey::{public_to_address, NetworkId, PlatformAddress, Public};
 use cstate::FindActionHandler;
-use ctypes::transaction::{Action, ShardTransaction as ShardTransactionType};
+use ctypes::transaction::{Action};
```
sgkim126:
```rust
use ctypes::transaction::Action;
```

rpc/src/v1/impls/chain.rs
```diff
@@ -14,36 +14,33 @@
...
 use ctypes::{BlockHash, BlockNumber, ShardId, Tracker, TxHash};
-use primitives::{Bytes as BytesArray, H160, H256};
+use primitives::{H256};
```
sgkim126:
```rust
use primitives::H256;
```

rpc/src/v1/impls/chain.rs
```diff
@@ -14,36 +14,33 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::convert::{TryFrom, TryInto};
+use std::convert::{TryFrom};
```
sgkim126:
```rust
use std::convert::TryFrom;
```

rpc/src/v1/impls/chain.rs
majecty: Thanks. I'll fix them.

majecty:
I fixed a merge conflict

PR #37, opened 2019-12-27
Introduce PreviousValidators and fix tendermint bugs

Remagpie:
The patches in https://github.com/CodeChain-io/foundry/pull/27 were originally ported from the CodeChain's snapshot branch, which includes CurrentValidators patch in CodeChain-io/codechain#1922. I found out that there was something wrong about the previous implementation, and introduced PreviousValidator state value to solve this.
It's almost the same as the CurrentValidators, except that it is updated with the values from CurrentValidators.
The fee calculation function is changed a bit to utilize this state value. Please take a look at it carefully.

Some other patches not related to the fee calculation are included too. A detailed explanation is written in the commit messages.

### DIFF
--- a/core/src/client/client.rs
+++ b/core/src/client/client.rs
@@ -481,6 +481,10 @@ impl StateInfo for Client {
 }
 
 impl EngineInfo for Client {
+    fn network_id(&self) -> NetworkId {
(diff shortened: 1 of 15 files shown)


sgkim126:
@HoOngEe please review it.

PR #83, opened 2020-01-06
Reorder imports

### LINKED ISSUE #41: Reorder imports
Currently, foundry group imports into three categories, but it's easy to make a mistake.
Let's reorder all imports without categorization.
https://gitter.im/CodeChain-io/codechain?at=5e084bfa8ba16b107ce94d12

junha1:
I implemented simple python script to apply this work on all bunch of .rs files.
https://gist.github.com/junha1/3ca19ab666140b527fc9f368f785d5e9
Fixes #41

### DIFF
--- a/core/src/client/importer.rs
+++ b/core/src/client/importer.rs
@@ -14,17 +14,6 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use std::borrow::Borrow;
-use std::collections::{HashMap, HashSet};
-use std::sync::Arc;
-
-use cio::IoChannel;
-use ctypes::header::Header;
-use ctypes::BlockHash;
-use kvdb::DBTransaction;
-use parking_lot::{Mutex, MutexGuard};
-use rlp::Encodable;
-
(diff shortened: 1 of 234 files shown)


majecty:
@junha1 Did you check the failed test case?

core/src/client/importer.rs
```diff
@@ -271,7 +269,6 @@ impl Importer {
             );
         })?;
 
-
```
sgkim126: It seems that your script removed some lines unrelated to this change.
Please check it before you apply this script to CodeChain.

sgkim126: I checked the scripts; there is no code to remove these kinds of lines.
I think it's your ide environment that removed these lines. Please check it.

sgkim126: By the way, why not
```python
for i in range(n - 2):
```
at https://gist.github.com/junha1/3ca19ab666140b527fc9f368f785d5e9#file-use-clauses_mering_script-py-L21-L22

junha1: @majecty timeout @sgkim126 Sorry I just checked it. I'll fix that.

PR #85, opened 2020-01-07
Add an RPC deleteAllPendingTransactions()

junha1:
This new RPC takes no argument and deletes all pending transactions(both current and future), returning nothing.

(See https://github.com/CodeChain-io/codechain/pull/1941)

### DIFF
--- a/test/src/e2e/mempool.test.ts
+++ b/test/src/e2e/mempool.test.ts
@@ -1,4 +1,4 @@
-// Copyright 2018-2019 Kodebox, Inc.
+// Copyright 2018-2020 Kodebox, Inc.
 // This file is part of CodeChain.
 //
 // This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


rpc/src/v1/impls/net.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2019 Kodebox, Inc.
+// Copyright 2018-2020 Kodebox, Inc.
```
majecty: Please revert this line.

test/src/e2e/mempool.test.ts
```diff
@@ -55,7 +55,6 @@ describe("Future queue", function() {
 
     it("all pending transactions must be mined", async function() {
         const seq = (await node.sdk.rpc.chain.getSeq(faucetAddress)) || 0;
-
```
majecty: Please add a new line here.
Please contain only needed changes in a commit.

junha1:
Commit message line wrapped

PR #89, opened 2020-01-07
Convert constants to discriminated enum

somniumism:
I implemented the Encodable/Decodable in the enum type so that
the RLP code will receive enum type directly.
The compiler will warn programmers if there are missing enum values in a
match statement by using discriminated enums instead of constant type.
Using an enum type rather than constant type provides more safety
against type error.


sgkim126:
@majecty Sorry. I misread that your updating branch as approval. I'll make a revert commit if you don't want to approve it.

majecty:
@sgkim126 It looks good to me!

PR #95, opened 2020-01-12
Remove parameters related to the removed Store/Remove transactions

### LINKED ISSUE #17: Remove Store and Remove Transaction
Anyone can save any data in CodeChain using the Store transaction. They can also remove it by using the Remove transaction.

In Foundry, that can be done using the module system. Let's remove them.

dynaxis:
~~Deprecate the relevant parameters and remove them also from the result of `chain_getMinTransactionFee`.~~

Remove parameters related to the removed Store/Remove transactions completely, as per the conclusion from discussion with @sgkim126.

Closes #17.


types/src/errors/runtime_error.rs
```diff
@@ -133,8 +130,9 @@ enum ErrorID {
     RegularKeyAlreadyInUseAsPlatform = 20,
     ScriptHashMismatch = 21,
     ScriptNotAllowed = 22,
-    TextNotExist = 23,
-    TextVerificationFail = 24,
+    /// Deprecated
```
sgkim126: I think we can remove the comments too.

types/src/common_params.rs
```diff
@@ -105,12 +107,15 @@ impl CommonParams {
     pub fn min_custom_transaction_cost(&self) -> u64 {
         self.min_custom_transaction_cost
     }
+    #[deprecated]
     pub fn min_store_transaction_cost(&self) -> u64 {
```
sgkim126: It seems that we can remove these functions instead of deprecating since the previous commit removed the related transaction.

types/src/errors/runtime_error.rs
dynaxis: As you see right above here, there are a few errors which got the exactly same treatment. If you are ok with me only removing the comments for Text related errors, I'll do that.
```
    /// Deprecated
    // INVALID_DECOMPOSED_INPUT = 13,
    // INVALID_DECOMPOSED_OUTPUT = 14,
```
To be sure, do you mean I need to remove the `/// Deprecated` line alone or all the three lines including it altogether?

sgkim126: All three lines.

types/src/common_params.rs
dynaxis: Here I also hope to remind you that `min_[de]compose_asset_cost` and the likes are not removed either. According to my examination over the codebase, they are not used elsewhere except related to RPC. The store/remove related functions are in the exactly same situation.

Whlie re-examining the relavent codes per your comment, I've found that we may remove these functions you mentioned if I do either of the following:

1. Remove the cost parameters completely from RPC and all the related e2e tests
1. Just break ties of those functions to RPC and remove the functions (simpler)

IMHO, the `min_[de]compose_asset_cost` things need the same treatment. So it was my proposal to treat them algether as a follow-up for this PR. But if you want me to take immediate action where possible, I hope to ask you of which way (1 or 2) to go for it.

sgkim126: How about removing the store/remove related features in this commit?
The compose/decompose related features will be removed by https://github.com/CodeChain-io/foundry/pull/11

dynaxis: I filed issue #96, but I can improve this PR to cover the issue #96 but only for the store/remove Tx-related codes. Is that what you want? If so, I am happy to do it, assuming the other entries are handled as per #11.

sgkim126: Yes, that's what I wanted. Thank you for doing it.

dynaxis:
I updated this PR with the relevant changes to completely remove the parameters from our codebase. But I keep failing on `start-dyn-val-2` e2e test, and couldn't know what's wrong and if it's anyhow related to the changes in this PR. Moreover I'm not sure my dev env at home is enough to complete the e2e test propertly, since they all have only 8GB RAM. Let's see how the result will be.

json/src/scheme/params.rs
```diff
@@ -93,15 +91,13 @@ mod tests {
...
-            "minStoreCost" : 17,
-            "minRemoveCost" : 18,
-            "minMintAssetCost" : 19,
-            "minTransferAssetCost" : 20,
-            "minChangeAssetSchemeCost" : 21,
-            "minComposeAssetCost" : 22,
-            "minDecomposeAssetCost" : 23,
-            "minUnwrapCccCost" : 24,
-            "minIncreaseAssetSupplyCost": 25,
+            "minMintAssetCost" : 17,
```
sgkim126: You don't need to change the values of other fields but it will not be matter, so I approved it.

dynaxis: Expect you would say so. ;)
I admit was a bit paranoid there.

dynaxis:
I just rebased this PR, since there was another PR merged in between. I don't know what is the usual workflow here when I get approval from one of two reviewers, one remaining, and need to rebase PR for merge. If what I'm doing is considered a bad practice, please enlighten me. Thanks.

majecty:
It is okay to rebase your PR branch on the master branch. Reviewers will re-check the PR.

Since the rebase did not change anything related to your PR and all the tests passed, I merged it.

PR #101, opened 2020-01-14
Calculate the transactions root with the transactions on the block

sgkim126:
Currently, the transactions root was the skewed merkle root of all
accumulated transactions from the genesis block. It changes the
definition of the transactions root to the skewed merkle root of
transactions in the block.

This change makes the node be able to verify a block's transactions
root without the parent header.

### DIFF
--- a/core/src/verification/verification.rs
+++ b/core/src/verification/verification.rs
@@ -21,6 +21,7 @@ use crate::consensus::CodeChainEngine;
 use crate::error::{BlockError, Error};
 use crate::transaction::{SignedTransaction, UnverifiedTransaction};
 use crate::views::BlockView;
+use ccrypto::BLAKE_NULL_RLP;
 use cmerkle::skewed_merkle_root;
 use ctypes::util::unexpected::{Mismatch, OutOfBounds};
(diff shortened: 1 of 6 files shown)


sgkim126:
@majecty Thank you. I fixed the typo.
And I added "This change makes the node be able to verify a block's transactions root without the parent header." at the end of the commit message.

sgkim126:
e2e test failed. I'll check it again.

core/src/verification/verification.rs
```diff
@@ -44,9 +45,15 @@ pub fn verify_block_basic(header: &Header, bytes: &[u8]) -> Result<(), Error> {
...
 
-    for t in body_rlp.iter().map(|rlp| rlp.as_val::<UnverifiedTransaction>()) {
-        t?.verify_basic()?;
-    }
+    let raw_transactions = body_rlp
+        .iter()
+        .map(|rlp| {
+            let tx = rlp.as_val::<UnverifiedTransaction>()?;
+            tx.verify_basic()?;
+            Ok(rlp.as_raw().to_vec())
```
sgkim126: I used the transaction hash in here; it should be the raw bytes of transactions.

PR #104, opened 2020-01-16
Do not prohibit headers whose body is malformed

sgkim126:
It depends on https://github.com/CodeChain-io/foundry/pull/101.


sgkim126:
I rebased it onto the master branch.

PR #106, opened 2020-01-16
Remove code related to reorganization

### LINKED ISSUE #31: Remove code related to reorganization

HoOngEe:
It closes #31


core/src/miner/mem_pool_types.rs
```diff
@@ -33,21 +33,17 @@ pub enum TxOrigin {
     Local,
     /// External transaction received from network
     External,
-    /// Transaction from retracted blocks
-    RetractedBlock,
```
sgkim126: I'm not sure about it.
What happened for the transactions in the rejected proposal block?

test/src/e2e.long/reward2.test.ts
```diff
@@ -51,149 +51,6 @@ describe("reward2", function() {
         ).to.deep.equal(new U64(50));
     }).timeout(30_000);
 
-    it("alice creates one block and bob creates two blocks in parallel. And then, sync", async function() {
```
sgkim126: I think it should be the first commit of this PR.

core/src/miner/mem_pool_types.rs
HoOngEe: Thank you, this change is not desirable. Retracted transactions by one-block retraction due to consensus failure (or by randomized-leader-election) should be considered.

test/src/e2e.long/reward2.test.ts
HoOngEe: I changed the order

HoOngEe: I changed the order

sgkim126:
@majecty Would you review tendermint related parts?

core/src/client/importer.rs
```diff
@@ -134,40 +134,24 @@ impl Importer {
...
+                set.insert(*hash);
+            });
+            set
         });
-
-        // Split to enacted retracted (using hashmap value)
-        let (enacted, retracted) = map.into_iter().partition(|&(_k, v)| v);
-        // And convert tuples to keys
-        (map_to_vec(enacted), map_to_vec(retracted))
+        set.into_iter().collect()
```
sgkim126: You can use `FromIterator` for it.

core/src/miner/mem_pool_types.rs
majecty: @sgkim126 and @HoOngEe What do you think about removing transactions from Memory pool only when the transactions are included in a finalized block?

Then we don't need to add transactions that was removed from the Memory pool to the Memory pool again.

core/src/client/importer.rs
```diff
@@ -134,40 +134,24 @@ impl Importer {
...
-        // will be in the hashmap
-        let map = import_results.iter().fold(HashMap::new(), |mut map, route| {
-            for hash in &route.enacted {
-                map.insert(*hash, true);
-            }
-            for hash in &route.retracted {
-                map.insert(*hash, false);
-            }
-            map
+    pub fn extract_enacted(&self, import_results: &[ImportRoute]) -> Vec<BlockHash> {
```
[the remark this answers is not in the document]
HoOngEe: Thank you for your advice!

core/src/client/importer.rs
HoOngEe: Is there any noticeable difference between `set.into_iter().collect()` and `Vec::from_iter(set.into_iter())`?

sgkim126: You can call `Vec::from_iter(set)` because `HashSet` implements`IntoIterator`.

core/src/miner/mem_pool_types.rs
HoOngEe: There are two methods removing transactions from the mem pool. One is `remove` and the other is `remove_old`. There is only one explicit call `remove` which removes invalid transactions(txs from malicious users or verification failed txs). Also there is one explicit call `remove_old` in `chain_new_blocks` which removes old (the norm of old here is the accounts' sequences in txs are old) transactions.
This `old` transactions are determined by chain's accounts' sequences in the latest block. The latest block fetches best block so the transactions in blocks rejected by consensus will not be removed from the mem pool.

core/src/client/importer.rs
HoOngEe: I modified the code

sgkim126:
I have a question about the commit titled "No longer consider retracted txs in db".
Doesn't we store the proposal block to prepare to reboot the node?
It seems that we should keep this code to handle the case that a proposal block is rejected, but I'm not sure. @majecty would you help me?

majecty:
@sgkim126 We don't need to consider the proposal block.
CodeChain does not remove transactions from the Memory pool when receiving a proposal block or creating a proposal block.
Transactions in the Memory pool removed only when the best block is changed.

PR #111, opened 2020-01-22
Remove simple_poa consensus

Byeongjee:
It resolves #111 
I removed simple_poa consensus since it's no longer needed.
Saluki chain which exploits simple_poa engine is also removed.


majecty:
Looks good to me. Let's wait the CI tests.

majecty:
Please use full sentences in the commit description.

junha1:
Good job

Byeongjee:
@sgkim126  @majecty  I changed a lot.
I removed force/mandatory resealing, and some trace messages which seem to be wrong.
Please review them carefully.

core/src/client/client.rs
```diff
@@ -76,7 +76,7 @@ pub struct Client {
 
     importer: Importer,
 
-    /// Timer for reseal_min_period/reseal_max_period on miner client
```
sgkim126: AFAIR, The tendermint consensus also uses reseal_max_period to prepare the situation there is no transactions in the block.
@majecty Would you check it?

majecty: Tendermint consensus always creates a block even though there is no transaction. So it is safe to remove it.

FYI, there is a separate timer in Tendermint code to sleep a few seconds when creating an empty block. However, the timer will be removed after the Tenderand is merged.

Byeongjee:
@majecty Travis CI test passed successfully on my local machine.

majecty:
@Byeongjee Please resolve the conflict.

PR #115, opened 2020-01-26
Re-send a body request when the body is corrupted

sgkim126:
Currently, Foundry doesn't mark the block as bad when the block is corrupted, but it doesn't re-request the body. This patch makes it re-send the request for the body.

~~It depends on https://github.com/CodeChain-io/foundry/pull/114.~~ #114 has been merged.

### DIFF
--- a/sync/src/block/downloader/body.rs
+++ b/sync/src/block/downloader/body.rs
@@ -53,13 +53,17 @@ impl BodyDownloader {
     }
 
     pub fn import_bodies(&mut self, hashes: Vec<BlockHash>, bodies: Vec<Vec<UnverifiedTransaction>>) {
+        let empty_targets: HashSet<_> = self.targets.iter().filter(|t| t.is_empty).map(|t| t.hash).collect();
         for (hash, body) in hashes.into_iter().zip(bodies) {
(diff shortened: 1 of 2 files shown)


sync/src/block/downloader/body.rs
```diff
@@ -55,11 +55,15 @@ impl BodyDownloader {
     pub fn import_bodies(&mut self, hashes: Vec<BlockHash>, bodies: Vec<Vec<UnverifiedTransaction>>) {
         for (hash, body) in hashes.into_iter().zip(bodies) {
             if self.downloading.remove(&hash) {
+                let target = self.targets.iter().find(|t| t.hash == hash).expect("Downloading target must exist");
```
majecty: I have a concern that this line of code will lookup n^2 of elements.
What do you think about creating `let target_map: HashMap<H256, Target>` before the for loop.

sgkim126: Looks good. I'll fix it.

sgkim126: Since it only checks whether the target is empty or not, I made a set of targets having an empty body.

PR #116, opened 2020-01-26
Refactor BodyDownloader

sgkim126:
There are five states for block hashes.
1. listed, but no request sent
2. sent the request
3. received the response
4. drained from the list
5. imported on the chain
    
Currently, these states are implicit with the combination of targets, downloading, and downloaded fields.
This patch merges downloading and downloaded fields, and make the states explicit.
There are four explicit states. The 5th state does not exist in BodyDownloader because it doesn't care about the imported blocks.


sync/src/block/downloader/body.rs
```diff
@@ -16,101 +16,142 @@
...
     pub fn re_request(&mut self, hash: BlockHash, remains: Vec<(BlockHash, Vec<UnverifiedTransaction>)>) {
+        let hashes = remains.iter().map(|(hash, _)| *hash);
         let mut new_targets = vec![hash];
-        new_targets.extend(remains.into_iter().map(|(hash, transactions)| {
-            self.downloaded.insert(hash, transactions);
-            hash
-        }));
+        new_targets.extend(hashes);
         new_targets.append(&mut self.targets);
         self.targets = new_targets;
```
majecty: Do we still add block hashes to the `self.target` here?
The reason why the previous `re_request` added block hashes to the `self.target` is that the `drain` function was removing block_hashes from the `self.target`.

Since the new `drain` function is not removing block hashes from the `self.target`. We must not add them to the `self.target` here.

sgkim126: You are right. It must not add the targets.
I'll fix it.

PR #117, opened 2020-01-28
Add RPCs for checking peer best blocks & target blocks

### LINKED ISSUE #71: Add RPCs for checking sync peers & sync target blocks

Byeongjee:
It resolves #71

### DIFF
--- a/spec/JSON-RPC.md
+++ b/spec/JSON-RPC.md
@@ -348,7 +348,8 @@ When `Transaction` is included in any response, there will be an additional fiel
  * [devel_startSealing](#devel_startsealing)
  * [devel_stopSealing](#devel_stopsealing)
  * [devel_getBlockSyncPeers](#devel_getblocksyncpeers)
-
+ * [devel_getPeerBestBlockHashes](#devel_getpeerbestblockhases)
+ * [devel_getTargetBlockHashes](#devel_gettargetblockhashes)
 
 # Specification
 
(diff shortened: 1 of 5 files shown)


sync/src/block/extension.rs
```diff
@@ -663,6 +663,16 @@ impl NetworkExtension<Event> for Extension {
                     channel.send(*peer).unwrap();
                 }
             }
+            Event::GetPeerBestBlockHashes(channel) => {
+                for header_downloader in self.header_downloaders.values() {
+                    channel.send(header_downloader.best_hash()).unwrap();
```
majecty: We need the information which peer has which best block hash.
Please pass the IP address of the peer also.

HoOngEe:
Is there any special reason to select `ipAddr` as a return value? I think ip address is not enough to be used in peer identification. `SocketAddr` seems to be great. @majecty what do you think of it?

majecty:
Aha, SocketAddr has a port number. I agree with @HoOngEe.

Byeongjee:
I changed IpAddr into SocketAddr and updated JSON-RPC documentation

spec/JSON-RPC.md
```diff
@@ -3063,6 +3064,70 @@ No parameters
...
+
+Get IP address and best block hash of each peer.
+
+### Params
+
+No parameters
+
+### Returns
+
+{ SocketAddress: `string`, BlockHash: `H256`}[]
```
majecty: This type does not match the example.
What you said:
```js
[ 
{ "SocketAddress": "xxx", "BlockHash": "yyy"},
{ "SocketAddress": "xxx2", "BlockHash": "yyy2"}, ...
]
```

However, in the example:
```js
[["1.2.3.4:3485", "0x56642f04d519ae3262c7ba6facf1c5b11450ebaeb7955337cfbc45420d573077"]]
```

Byeongjee: Okay.  Should I introduce new type such as SocketAddressAndHash, or tuple type is sufficient?

majecty: IMO, tuple type is sufficient.

Byeongjee: Then the return type should be [ 'string', 'H256'][]?

majecty: It looks good to me.

Byeongjee:
I fixed both license part and return type.

rpc/src/v1/traits/chain.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2019 Kodebox, Inc.
+// Copyright 2018-2020 Kodebox, Inc.
```
majecty: Please update the copyrights in the files that you changed.

Byeongjee: Sorry. I changed the copyright of wrong file. I pushed the fixed version.

Byeongjee:
There was timeout exceptions in some tests, but they passed successfully in my local machine. @majecty

PR #118, opened 2020-01-29
Fix auto-self-nonimante

majecty:


foundry/auto_self_nominate.rs
```diff
@@ -134,8 +134,13 @@ impl AutoSelfNomination {
         let candidate = Candidates::load_from_state(&state).unwrap();
         if candidate.get_candidate(&address).is_some() {
             let candidate_need_nomination = candidate.get_candidate(&address).unwrap();
-            if candidate_need_nomination.nomination_ends_at <= current_term + NEED_NOMINATION_UNDER_TERM_LEFT {
-                cdebug!(ENGINE, "No need self nominate");
+            if candidate_need_nomination.nomination_ends_at + NEED_NOMINATION_UNDER_TERM_LEFT <= current_term {
```
majecty: We missed this condition.

MSNTCS: Yes, I think it should be reversed.

PR #120, opened 2020-01-29
 Remove TxTimlock from the mempool

majecty:
I updated the PR #12.

### DIFF
--- a/core/src/miner/mem_pool.rs
+++ b/core/src/miner/mem_pool.rs
@@ -267,10 +267,7 @@ impl MemPool {
 
             let id = self.next_transaction_id;
             self.next_transaction_id += 1;
-            let item = MemPoolItem::new(tx, origin, inserted_block_number, inserted_timestamp, id, TxTimelock {
-                block: None,
-                timestamp: None,
-            });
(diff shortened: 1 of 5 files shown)


core/src/miner/mem_pool.rs
```diff
@@ -601,61 +587,13 @@ impl MemPool {
 
     /// Checks the timelock of transactions starting from `start_seq`.
```
sgkim126: I think we should change the comment.

core/src/miner/mem_pool.rs
```diff
@@ -601,61 +587,13 @@ impl MemPool {
...
-    /// If there's no transaction which should wait timelock, returns None.
-    fn check_new_transactions(
-        &self,
-        public: Public,
-        seqs: &[u64],
-        next_seq: u64,
-        current_block_number: PoolingInstant,
-        current_timestamp: u64,
-    ) -> Option<u64> {
+    fn check_transactions(&self, public: Public, start_seq: u64) -> u64 {
```
sgkim126: And I think renaming this method to `next_seq_of_queued` would be good.

majecty:
@sgkim126 I updated this PR.

majecty:
I fixed conflicts.

PR #121, opened 2020-01-29
Remove needless weight data from Validator

HoOngEe:
In randomized leader election environment, Foundry doesn't need
to manage weight values. Previously, it was utilized to guarantee
stake-proportional leader election but now the property comes from
the nature of the probability model. 
Since this data type change can affect to light client implementation,
I think it's better to be merged earlier.


sgkim126:
Does the RLE treat all validators have the same vote?
What does happen if the major number of validators agreed, but the major amount of stakes does not?

HoOngEe:
`Validator` struct has both `weight` and `delegation` fields. When the struct is initialized, both are set to one equal value. `weight` value decreases as rounds proceed and `Validators` were re-sorted by `weight` values. It wasn't used to measure voting power but just helped leader elections in weighted round-robin fashion. Still voting power can be measured using the `delegation` field.

sgkim126:
@HoOngEe Would you check e2e tests?

HoOngEe:
I'm sorry, I was day off yesterday. I'll check it.

HoOngEe:
I'm sorry, this commit can not be solely applied to the master. If the weight data is removed, the round-robin fashioned election will not work so that almost all the tests related to dynamic validator will fail because there is only one fixed validator.

HoOngEe:
I think this issue can only be merged into master when Tenderand becomes the main consensus of Foundry unavoidably.

PR #122, opened 2020-01-29
Migrate IBC transactions to use TopLevel transaction

majecty:
The previous ics-poc implementation was using the custom action handler. Since Seulgi will remove the custom action handler, I'm porting them to use TopLevel transactions.


majecty:
@junha1 Could you check this PR?
You don't need to read code line by line.
Just check which files are changed and which struct is created and removed.

PR #127, opened 2020-01-30
Add an RPC getting minimum fees confiugred by each machine

### LINKED ISSUE #56: Need a new RPC getting minimum fees configured by each machine
After the CodeChain-io/codechain#1827, CodeChain has two sorts of minimum fees. One is defined by the network, the other is defined by each machine. For users not to confuse these new fee thresholds with common parameters, CodeChain needs a new RPC getting minimum fees configured by each machine.

Byeongjee:
It closes #56

### DIFF
--- a/core/src/miner/mem_pool_types.rs
+++ b/core/src/miner/mem_pool_types.rs
@@ -421,18 +421,18 @@ pub struct AccountDetails {
 #[derive(Default, Clone, Copy, Debug, PartialEq)]
 /// Minimum fee thresholds defined not by network but by Mempool
 pub struct MemPoolMinFees {
-    min_pay_transaction_cost: u64,
-    min_set_regular_key_transaction_cost: u64,
-    min_create_shard_transaction_cost: u64,
-    min_set_shard_owners_transaction_cost: u64,
-    min_set_shard_users_transaction_cost: u64,
-    min_wrap_ccc_transaction_cost: u64,
-    min_custom_transaction_cost: u64,
-    min_asset_mint_cost: u64,
-    min_asset_transfer_cost: u64,
-    min_asset_scheme_change_cost: u64,
-    min_asset_supply_increase_cost: u64,
-    min_asset_unwrap_ccc_cost: u64,
+    pub min_pay_transaction_cost: u64,
+    pub min_set_regular_key_transaction_cost: u64,
+    pub min_create_shard_transaction_cost: u64,
+    pub min_set_shard_owners_transaction_cost: u64,
+    pub min_set_shard_users_transaction_cost: u64,
+    pub min_wrap_ccc_transaction_cost: u64,
+    pub min_custom_transaction_cost: u64,
+    pub min_asset_mint_cost: u64,
+    pub min_asset_transfer_cost: u64,
+    pub min_asset_scheme_change_cost: u64,
+    pub min_asset_supply_increase_cost: u64,
+    pub min_asset_unwrap_ccc_cost: u64,
 }
 
 impl MemPoolMinFees {
--- a/core/src/miner/mem_pool_types.rs
+++ b/core/src/miner/mem_pool_types.rs
@@ -420,7 +420,7 @@ pub struct AccountDetails {
 
 #[derive(Default, Clone, Copy, Debug, PartialEq)]
 /// Minimum fee thresholds defined not by network but by Mempool
-pub struct MemPoolFees {
+pub struct MemPoolMinFees {
     min_pay_transaction_cost: u64,
     min_set_regular_key_transaction_cost: u64,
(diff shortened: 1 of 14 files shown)


core/Cargo.toml
```diff
@@ -35,6 +35,8 @@ rlp_compress = { git = "https://github.com/CodeChain-io/rlp.git", version = "0.2
 rlp_derive = { git = "https://github.com/CodeChain-io/rlp.git", version = "0.2" }
 snap = "0.2"
 table = { path = "../util/table" }
+serde = "1.0"
+serde_derive = "1.0"
```
majecty: We aren't using serde in core crate.
Since serializing a data structure can be different by the context, we don't provide a default serialization for the core data types.
Please create a new type in rpc/src/v1/types directory and create serialization code for the type.

spec/JSON-RPC.md
```diff
@@ -1931,6 +1932,63 @@ curl \
...
+    "minAssetTransferCost":`number`,
+    "minAssetUnwrapCccCost":`number`,
+    "minCreateShardTransactionCost":`number`,
+    "minCustomTransactionCost":`number`,
+    "minPayTransactionCost":`number`,
+    "minSetRegularKeyTransactionCost":`number`,
+    "minSetShardOwnersTransactionCost":`number`,
+    "minSetShardUsersTransactionCost":`number`,
+    "minWrapCccTransactionCost":`number`
+    }
```
majecty: You wrote:
```
    }
```

Fix it like:
```
}
```

spec/JSON-RPC.md
```diff
@@ -1931,6 +1932,63 @@ curl \
 
 [Back to **List of methods**](#list-of-methods)
 
+## mempool_getMinimumFees
```
majecty: The RPC method seems to general.
Since we have two types of minimum fee, we need a verbose RPC name.

core/Cargo.toml
Byeongjee: Which is preferred, to use serde remote derive (https://serde.rs/remote-derive.html), or to create a new type with the same name?

majecty: Let's make it consistent with other types in the rpc trait. You can find them in the rpc/src/v1/types directory. Since they are creating a new type with the same name, let's use the method.

core/src/miner/mem_pool_types.rs
```diff
@@ -429,22 +429,22 @@ pub struct AccountDetails {
...
-    min_set_shard_owners_transaction_cost: u64,
-    min_set_shard_users_transaction_cost: u64,
-    min_wrap_ccc_transaction_cost: u64,
-    min_custom_transaction_cost: u64,
-    min_asset_mint_cost: u64,
-    min_asset_transfer_cost: u64,
-    min_asset_scheme_change_cost: u64,
-    min_asset_supply_increase_cost: u64,
-    min_asset_unwrap_ccc_cost: u64,
+pub struct MemPoolMinFees {
```
majecty: @Byeongjee Can you separate this commit into the "Renaming MemPoolFees" commit and the "Creating an RPC" commit? If it is hard, it's okay to maintain the commit since this commit is small.

It is a good practice to create a commit for refactoring. When we review a commit, we focus on the different aspects when reviewing a refactoring commit and feature commit.
When we review a refactoring commit, we focus on that the behavior is changed or not changed. A refactoring commit should not change behavior. When we review a feature commit, we focus on whether the code is changed as expected.

rpc/src/v1/types/mod.rs
```diff
@@ -16,17 +16,18 @@
...
 mod work;
 
 pub use self::action::{Action, ActionWithTracker};
 pub use self::block::Block;
 pub use self::block::BlockNumberAndHash;
+pub use self::mem_pool::MemPoolMinFees;
 pub use self::transaction::{PendingTransactions, Transaction};
 pub use self::unsigned_transaction::UnsignedTransaction;
 pub use self::work::Work;
-
```
majecty: https://github.com/kodebox-io/kodebox-internal/wiki/Tips-for-Developers#selectively-stage-your-work

spec/JSON-RPC.md
```diff
@@ -1931,6 +1932,63 @@ curl \
...
+  "minCreateShardTransactionCost":`number`,
+  "minCustomTransactionCost":`number`,
+  "minPayTransactionCost":`number`,
+  "minSetRegularKeyTransactionCost":`number`,
+  "minSetShardOwnersTransactionCost":`number`,
+  "minSetShardUsersTransactionCost":`number`,
+  "minWrapCccTransactionCost":`number`
+}
+
+Error: `Invalid params`
```
majecty: This RPC does not receive any params. When is the `Invalid params` returned?

PR #131, opened 2020-01-30
Do not consider reorganization in the Foundry

HoOngEe:


HoOngEe:
I couldn't find any special optimization point related to removing reorganization. I think it's better to revisit this part after removing the concept of score.

sync/src/block/extension.rs
```diff
@@ -792,24 +792,12 @@ impl Extension {
...
-                };
-                self.client.block(&block_id)
-            })
-            .take_while(Option::is_some)
-            .map(|block| block.expect("take_while guarantees existance of item").header().decode())
+        // In Tendermint the best proposal header can only be fetched by the block hash
+        let headers = (start_number..start_number + max_count)
+            .take_while(|number| *number < best_proposal_header.number())
+            .map(BlockId::Number)
+            .chain(vec![best_proposal_header.hash().into()].into_iter())
```
sgkim126: You can use [`once`](https://doc.rust-lang.org/std/iter/fn.once.html).

HoOngEe: Thank you I modified it.

HoOngEe:
@sgkim126 I modified the code to get expected result as `.expect` message says. I didn't scrutinize all the logic but I realized from test failures that in some states client can return `block_header()` successfully for the `best_proposal_block_header` but cannot return `block()`. Previously, the corresponding panic was shielded by `.take_while(Option::is_some)`.
I thought the previous behavior is weird because the code get header by indirect way: get block first and look up header from the block. Considering the synchronicity of header information, I think the changed code is desirable.

sgkim126:
Hi @majecty,
I have a question about the best block proposal header.
Is the only case that there is the best block proposal header without block body the node received the proposal, but it's not committed yet?

sync/src/block/extension.rs
```diff
@@ -797,24 +798,12 @@ impl Extension {
...
-                self.client.block(&block_id)
-            })
-            .take_while(Option::is_some)
-            .map(|block| block.expect("take_while guarantees existance of item").header().decode())
+        // In Tendermint the best proposal header can only be fetched by the block hash
+        let headers = (start_number..start_number + max_count)
+            .take_while(|number| *number < best_proposal_header.number())
+            .map(BlockId::Number)
+            .chain(once(best_proposal_header.hash().into()))
+            .map(|block_id| self.client.block_header(&block_id).expect("Only valid block ids are selected").decode())
```
sgkim126: IMO, it should not.
The node should response headers which also has the body.
It seems that the previous code represents it.

majecty:
@sgkim126 
> Is the only case that there is the best block proposal header without block body the node received the proposal, but it's not committed yet?

There is a case when there are no bodies for headers, but it is not related to the Tendermint.
1. If a node received a proposal in Tendermint, there is always body for it.
2. When a node received headers from the sync extension, It may not have bodies for them.

sync/src/block/extension.rs
HoOngEe: Ok, I'll fit into the previous expected behavior.

PR #135, opened 2020-01-31
Remove duplicated TermInfo

majecty:


sgkim126:
It doesn't need to wait for e2e tests because this PR just removes a generic parameter constraint.
I checked it's compiled, I'll merge it.

PR #140, opened 2020-02-06
Reorder state_root and transactions_root

junha1:
This refactoring reorders the order of two fields in 'code', which was inconsistent.
RLP encoding and other run-time properties have not been changed.

### DIFF
--- a/core/src/scheme/genesis.rs
+++ b/core/src/scheme/genesis.rs
@@ -33,10 +33,10 @@ pub struct Genesis {
     pub timestamp: u64,
     /// Parent hash.
     pub parent_hash: BlockHash,
-    /// Transactions root.
-    pub transactions_root: H256,
     /// State root.
     pub state_root: Option<H256>,
+    /// Transactions root.
+    pub transactions_root: H256,
     /// The genesis block's extra data field.
     pub extra_data: Bytes,
 }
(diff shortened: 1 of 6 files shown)


junha1:
Oh, I found that other fields (parent_hash, author, timestamp... ) are all inconsistent too.

sgkim126:
What does inconsistent mean? There is no logical reason for ordering the fields.

junha1:
@sgkim126 Inconsistency with the order of the RLP encoding

sgkim126:
I will not disagree with this commit, but I do not agree with it either.
There is no reason to match the order of fields in rust code and rlp encoding.

junha1:
@sgkim126 After short discussion with @kseo, there was no reason of such inconsistent but no reason either to fix it. Furthermore this commit itself doesn't fix all inconsistency. I decided to just let them be and **close** this PR. However this commit is still helpful, since I pointed out all places to be fixed whenever someone wants to manage the block header's field.

sgkim126:
The logical correlation decides the field order in code, but the compatibility is more important in the order of RLp encoding. You will meet the situation that cannot be matched when you are updating schemas. Even some field exists only on code or is represented in rlp encoding.
So generally speaking, you cannot make them coherence.

junha1:
@sgkim126 Ok, I agree that

Issue #38, opened 2019-12-29
Change stake transactions to first-class transactions instead of custom transactions


sgkim126:
CodeChain claims to be a pluggable consensus engine so the transactions related to a consensus are custom transactions, but Foundry doesn't need to do it since it only provides tendermint consensus.

PR #149, opened 2020-02-09
Remove Custom transactions and make the stake transactions top level

sgkim126:
It closes https://github.com/CodeChain-io/foundry/issues/38


sgkim126:
It depends on https://github.com/CodeChain-io/foundry/pull/291

core/src/miner/mem_pool_types.rs
```diff
@@ -433,15 +433,13 @@ impl MemPoolMinFees {
...
-            Action::Custom {
-                ..
-            } => self.min_custom_transaction_cost,
             Action::ShardStore {
                 ..
             } => {
                 // FIXME
                 0
             }
+            _ => self.min_custom_transaction_cost,
```
majecty: How about creating match cases for each Actions Instead of using `_` pattern?
If we use `_` pattern in our code, it is hard to find the match expression when we add a new enum variant in the Action.

However, if this code will be removed and all transactions will be moved to the Module level, it's good to use `_` pattern here for temporary.

core/src/codechain_machine.rs
```diff
@@ -92,15 +92,13 @@ impl CodeChainMachine {
...
-            Action::Custom {
-                ..
-            } => params.min_custom_transaction_cost(),
             Action::ShardStore {
                 ..
             } => {
                 // FIXME
                 0
             }
+            _ => params.min_custom_transaction_cost(),
```
majecty: Will you change the name `min_custom_transaction_cost` later?

core/src/miner/mem_pool_types.rs
sgkim126: https://github.com/CodeChain-io/foundry/pull/292 will remove the minimum cost. I'll rebase it on #292.

core/src/codechain_machine.rs
sgkim126: ditto.

PR #130, opened 2020-01-30
Adapt e2e tests to the new foundry-rpc

MSNTCS:

PR #153, opened 2020-02-10
Add a new field 'next validator set hash' in the header

junha1:
I will add unit/e2e tests after https://github.com/CodeChain-io/foundry/pull/130 and reviews


types/src/header.rs
```diff
@@ -50,6 +51,8 @@ pub struct Header {
     transactions_root: H256,
     /// State root.
     state_root: H256,
+    /// Validator set root. Note that this is for the next block.
+    validator_set_root: H256,
```
majecty: Why did you choose the term "root"?
Is the validator set is stored in a tree?

majecty: What do you think about adding the "next_" prefix to the field name?
When I first read the "validator_set_root", it seems that it is the current block's validator set.

junha1: Ok then I'll rename it as `next_validator_set_hash` after your overall review

core/src/consensus/stake/action_data.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2019 Kodebox, Inc.
```
sgkim126: Please squash this commit to the preceding commit.

types/src/validator_set.rs
```diff
@@ -0,0 +1,57 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+use ccrypto::blake256;
+use ckey::Public;
+use primitives::H256;
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+
+// It will be hashed in the header.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct CompactValidatorSet(pub Vec<(Public, u64)>);
```
sgkim126: Why do we need this structure?
Why now implementing hash function instead of create_compact_validator_set for NextValidator?

majecty: I suggested to Junha to add this struct in the "types/src" directory.
If someone wants to know what is the "validator_set_root" in the header, reading this struct will help them.

junha1: Because it will be also passed to the light client for verification. Light client only needs that compact data of the validator set.

types/src/validator_set.rs
```diff
@@ -0,0 +1,57 @@
...
+            })
+        }
+        let mut vec = Vec::with_capacity(item_count / 2);
+        // TODO: Optimzie the below code
+        for i in 0..(item_count / 2) {
+            vec.push((rlp.val_at(i * 2)?, rlp.val_at(i * 2 + 1)?));
+        }
+        Ok(Self(vec))
+    }
+}
```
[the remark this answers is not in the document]
junha1: Unit test added. Check the same file `types/src/validator_set.rs` please.

Cargo.lock
```diff
@@ -540,6 +540,7 @@ dependencies = [
  "codechain-json",
  "codechain-key",
  "primitives",
+ "rand 0.6.1",
```
majecty: We should be careful when we add a dependency in Foundry.
Please discuss with me before adding a new package in the Foundry.

junha1: Ok I'll be careful next time

junha1: I think they're quite similar; Use `rand`, with constant seed and iterate.

majecty: Thanks. Let's use it.

core/src/block.rs
```diff
@@ -232,6 +233,12 @@ impl<'x> OpenBlock<'x> {
             e
         })?;
         self.block.header.set_state_root(state_root);
+
+        // Note : Does state always hold validator set?
```
majecty: Could you change this to "FIXME" like the below description?

> NextValidators do not exist in the static validator set.
This code does not work expected when using the static validator set.
Since we will change Foundry to use only the dynamic validator set, we don't add code that handles the static validator set.
After changing Foundry to use only the dynamic validator set, we should remove this comment.

junha1: Sounds good but once I tested this, it seems that State always carries validator set as an empty list in case of static mode. (I'm not sure)

junha1: Technically it doesn't cause a error even now since we use `unwrap_or_default()` in `load_from_state()`. 
So just simple comment will be ok.

> `NextValidators` entry doesn't exit in the state during static-validator-set-mode.
>  It doesn't cause direct error since we use `unwrap_or_default()` here, but should be aware of such problem
>  Remove this comment after we completely omit static-validator-set-mode

junha1: @majecty updated.

core/src/block.rs
```diff
@@ -232,6 +233,16 @@ impl<'x> OpenBlock<'x> {
             e
         })?;
         self.block.header.set_state_root(state_root);
+
+        /*
+        FIXME: NextValidators entry doesn't exit in the state during static-validator-set-mode.
```
majecty: exit- > exist?
Please review English grammar before updating PR.

junha1: @majecty typo fixed

majecty:
@sgkim126 I want to merge this PR.

PR #154, opened 2020-02-10
Replace ECDSA and Schnorr with Ed25519

### LINKED ISSUE #146: Remove SchnorrSignature
Remove `SchnorrSignature` from `key` package and replace it with `Ed25519Signature`.

HoOngEe:
It resolves #146


HoOngEe:
In this pr the e2e test cannot fully covered because this change requires SDK changes.

key/src/ed25519_private.rs
```diff
@@ -0,0 +1,99 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::ops::Deref;
+use std::str::FromStr;
+
```
sgkim126: Please remove the blank lines.
https://github.com/CodeChain-io/codechain/issues/1940

key/src/ed25519_private.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+impl From<SecretKey> for Private {
+    fn from(k: SecretKey) -> Self {
+        Private(k)
+    }
+}
+
+impl Encodable for Private {
```
sgkim126: Please add a test for rlp encoding.

key/src/ed25519_public.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl From<PublicKey> for Public {
+    fn from(k: PublicKey) -> Self {
+        Public(k)
+    }
+}
+
+impl Encodable for Public {
+    fn rlp_append(&self, s: &mut RlpStream) {
+        s.append_single_value(&self.as_ref());
```
sgkim126: Why does it use `append_single_value`, while you used `append` for `Private`?

core/src/account_provider.rs
```diff
@@ -264,7 +264,7 @@ mod tests {
     fn unlock_account_temp() {
         let kp = Random.generate().unwrap();
         let ap = AccountProvider::transient_provider();
-        assert!(ap.insert_account(*kp.private(), &"test".into()).is_ok());
```
sgkim126: Why do you prohibit copy?

key/src/lib.rs
```diff
@@ -26,52 +26,41 @@ extern crate rand_xorshift;
...
 extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
+extern crate sodiumoxide;
 
 mod address;
-mod ecdsa;
+mod ed25519;
+mod ed25519_private;
```
sgkim126: How about move `ed25519_private` and`ed25519_public` under the `ed25519` module?

key/src/ed25519_private.rs
HoOngEe: Thank you, I'll keep it in mind.

key/src/ed25519_public.rs
HoOngEe: Sorry, I should have used append_single_value for `Private`.
When I used `append` for the slice, I observed that the encoded `Public` is not separated as an independent list element in the bigger encoded structs.

HoOngEe: Even, I don't have to implement `Encodable` and `Decodable` for `Private`. I'll modify it.

core/src/account_provider.rs
HoOngEe: It is a very low-level security concern and `Rust` can safely dispose private information by implementing memory clear logic in `Drop` trait. However, the more conservative way is prohibiting copy behavior of private information. Previously, private information was stored in `H256` type and the `DecryptedAccount` which possibly contains private key info implements memory-clearing `Drop` trait.
However, I could not sure about other usages are desirable because `H256` was used just for hash information also. And this change did not require many changes. What do you think of it?

key/src/lib.rs
HoOngEe: I made `ed25519` module and put them into it.

core/src/account_provider.rs
majecty: @HoOngEe How about adding summary of this description in the commit message?

core/src/account_provider.rs
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use ckey::{public_to_address, Address, Error as KeyError, Generator, KeyPair, Password, Private, Public, Random};
+use ckey::{public_to_address, Address, Error as KeyError, Generator, Password, Private, Public, Random};
```
majecty: Great!
I'm happy to see this commit.

key/src/exchange.rs
```diff
@@ -15,23 +15,34 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 use crate::{Error, Private, Public, SharedSecret};
-use sodiumoxide::crypto::kx::server_session_keys;
+use sodiumoxide::crypto::kx::{client_session_keys, server_session_keys};
 use std::convert::TryInto;
 
-pub fn exchange(other_public: &Public, my_private: &Private) -> Result<SharedSecret, Error> {
-    let my_public = &my_private.public_key();
+pub fn server_session_key(client_public: &Public, server_private: &Private) -> Result<SharedSecret, Error> {
```
majecty: @HoOngEe I don't know what is the server_session_key and client_session_key, is it related to the Diffie-Hellman key exchange? Could you give me some material?

HoOngEe:
The key exchange function supported by `sodiumoxide` operates like follows:
It calculate a key by ECDH: q = (local_private_key) * (remote_public_key). In `client_session_key` 
, local means the client side and remote means the server side and vice versa. Then it hashes (q || client_public_key || server_public_key) into 64 bytes. It splits the result as tuple. The `server_session_key` returns (hash[0:32], hash[33:64]) and `client_session_key` returns (hash[33:64], hash[0:32]). If we want a single shared secret, we have to use the Fst(tuple_resulted_from_sever_session_key) == Snd(tuple_resulted_from_client_session_key) or the other.

key/src/exchange.rs
HoOngEe: https://github.com/CodeChain-io/foundry/pull/154#issuecomment-586122289

core/src/consensus/stake/actions.rs
```diff
@@ -127,20 +143,24 @@ impl Action {
...
                 let action = Action::ChangeParams {
                     metadata_seq: *metadata_seq,
                     params: params.clone(),
-                    signatures: vec![],
+                    approvals: vec![],
                 };
                 let encoded_action = H256::blake(rlp::encode(&action));
-                for signature in signatures {
+                for approval in approvals {
                     // XXX: Signature recovery is an expensive job. Should we do it twice?
```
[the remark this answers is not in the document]
HoOngEe: Thank you I removed the comment

key/src/ed25519_private.rs
HoOngEe: The `Encodable` for `Private` is removed

core/src/account_provider.rs
sgkim126: I'm not sure the purpose yet. Prohibiting a copy of private data is not a way to keep data privately. I can agree that H256 seems to general and you want to replace it with `SecretKey`, but it doesn't mean that SecretKey is safer because it doesn't provide `Copy`.

key/src/lib.rs
```diff
@@ -31,6 +31,7 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
+extern crate sodiumoxide;
```
sgkim126: You don't need to add extern crate statements anymore since 2018 edition.

HoOngEe: Thank you, I studied the comment on it after you pointed out.

key/src/ed25519/ed25519_private.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 Kodebox, Inc.
```
sgkim126: I think this can be renamed to private.rs since it's under the ed25519 module.

key/src/ed25519/ed25519_public.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright 2020 Kodebox, Inc.
```
sgkim126: Same as the above.

key/src/ed25519/ed25519_signature.rs
```diff
@@ -0,0 +1,121 @@
+// Copyright 2018-2020 Kodebox, Inc.
```
sgkim126: ditto.

core/src/account_provider.rs
HoOngEe: Oh my god. I misunderstood the types. The actual situation is: `SecretKey` by `sodiumoxide` implemented memory clearing `Drop` trait inside so the `Copy` marker is incompatible. Therefore, the `Private` which is a wrapper for `SecretKey` cannot mark `Copy` on it. In the sense, this change is inevitable. Is there some use cases need to allow `Copy` behavior?

key/src/ed25519/ed25519_private.rs
HoOngEe: Ok, that's more reasonable. I changed it.

core/src/account_provider.rs
sgkim126: No, we don't need `Copy`, we can use `Clone` instead of it.

sgkim126: Please change the commit message.

key/src/ed25519/private.rs
```diff
@@ -0,0 +1,73 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::Public;
+use primitives::H512;
+use sodiumoxide::crypto::sign::{gen_keypair, SecretKey, SECRETKEYBYTES};
+use std::ops::Deref;
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Private(SecretKey);
```
sgkim126: I think it's better to add a comment that `SecretKey` clears memory when it is dropped to inform others who want to replace the underlying type.

key/src/ed25519/private.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl FromStr for Private {
+    type Err = crate::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let for_slice = H512::from_str(s).map_err(|e| crate::Error::Custom(format!("{:?}", e)))?;
+        Ok(SecretKey::from_slice(&for_slice).expect("H512 has length 64").into())
+    }
+}
+
+impl Default for Private {
```
sgkim126: The only use case of `Default` of `Private` is `Drop` of `DecrptedAccount`, but it can be removed because `SecretKey` does the role of `Drop` of `DecrptedAccount`.

foundry/subcommand/convert_command.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Kodebox, Inc.
+// Copyright 2020 Kodebox, Inc.
```
sgkim126: 2019-2020

key/src/ed25519/private.rs
HoOngEe: Ok, I added a comment like "The inner type SecretKey clears its memory when it is dropped"

key/src/ed25519/private.rs
HoOngEe: Ok that's great

key/src/ed25519/public.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn from_slice(slice: &[u8]) -> Option<Self> {
+        PublicKey::from_slice(slice).map(Self)
+    }
+
+    pub fn is_zero(&self) -> bool {
+        self.as_ref() == [0; PUBLICKEYBYTES]
+    }
+}
+
+impl Default for Public {
```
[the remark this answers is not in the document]
HoOngEe: Thank you for your scrutiny I removed it

key/src/ed25519/public.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn default() -> Self {
+        Public(PublicKey::from_slice(&[0; PUBLICKEYBYTES]).expect("Slice from an array with proper length never fails"))
+    }
+}
+
+impl From<u64> for Public {
```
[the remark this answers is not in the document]
HoOngEe: I tried to add `#[cfg(test)]` to the implementation. However, conditional complied test implementations were invisible in other crates. I hoped to do so but it seems no sharp solution is available.

key/src/lib.rs
```diff
@@ -31,17 +31,16 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
-extern crate sodiumoxide;
```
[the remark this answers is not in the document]
HoOngEe: Ok, I moved the change to the previous commit

foundry/subcommand/convert_command.rs
HoOngEe: Thanks, fixed.

core/src/account_provider.rs
HoOngEe: I modified the commit message to be correct.

sgkim126:
@HoOngEe Please check the e2e tests. It seems that some RLP encodings are wrong.

key/src/ed25519/public.rs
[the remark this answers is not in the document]
sgkim126: Okay. It seems that we should leave it or replace the test code.
Let's leave it for now.

HoOngEe:
https://github.com/CodeChain-io/foundry/pull/154#issuecomment-585041845
To pass e2e test, it requires ed25519 JavaScript implementation. What is the recommended plan?

sgkim126:
@majecty @HoOngEe @kseo 
IMO, to skip tests will be dangerous since we have a plan to change many things.
How about finding the ed25519 implementation of JavaScript at this time?

HoOngEe:
@sgkim126 We can utilize the https://github.com/jedisct1/libsodium.js/. After the sdk code moves into the foundry, @somniumism will introduce javascript `ed25519`.

sgkim126:
It seems that the test still fails. @HoOngEe would you check it?

HoOngEe:
@sgkim126 Sorry, it's not done yet. I'll finish it by today.

HoOngEe:
I'm done until the tests triggered by `yarn start-short`. 
The remaining tests triggered by `yarn start-long` need a `derive` function to share secret keys in `p2pLayer`. The derive is previously just an ECDH. We can implement the counterpart with `tweetnacl` which is a libsodium wrapper. I think it's better to support the functionality in `foundry-primitives` for general purposes.
I also found some mishandled rejected expectations. I'll fix them in another pull request.

HoOngEe:
I checked all the tests including dynamic validator tests pass in my local environment except the `snapshot` test in the `dyn-val-2`. It also fails in the master branch and it should be fixed in another pr.

HoOngEe:
Unlike the first draft, I added two modules in the `key` crate.
1. Ed25519 keys and KeyPair for signatures
2. X25519 keys and KeyPair for key exchanges in p2p connection.

`key` crate exposes two key types with the prefixes `Ed25519` and `X25519`.
And each module imports such prefixed key types without prefix. For example, network related modules import` X25519KeyPair ` as `KeyPair` and other modules import  `Ed25519KeyPair` as `KeyPair`.

HoOngEe:
However, it can be tested successfully with `foundry-primitives` version `0.2.0` after removing the concept of `Asset` in the test framwork. I'll make a pr for it.

test/src/stakeholder/actions/reportDoubleVote.ts
```diff
@@ -46,16 +46,16 @@ export default class ReportDoubleVote {
...
+    // public criminal(): H512 {
+    //     const { message1 } = this;
+    //     const digest = blake256(RLP.encode(messageToEncodeObject(message1)[0]));
+    //     const signature = message1.signature.toString();
+    //     const pubkey = recoverSchnorr(digest, {
+    //         r: signature.slice(0, 64),
+    //         s: signature.slice(64)
+    //     });
+    //     return H512.ensure(pubkey);
+    // }
```
majecty: If this code is not used, let's remove it.

HoOngEe: Thank you, I removed it. I'll force-push after your review ends

majecty: I reviewed it. Other code looks good to me.

HoOngEe: Ok, I applied the change

test/src/e2e.dynval/1/doubleVote.test.ts
```diff
@@ -73,7 +73,7 @@ describe("Double vote detection", function() {
             aliceNode.testFramework
         );
         expect(banned.map(b => b.toString())).to.include(
-            betty.platformAddress.toString()
+            betty.address.toString()
```
sgkim126: Is the platform address just renamed to address, or was the type changed?

sgkim126: If the type is changed, we don't use the platform address anymore?

HoOngEe: It was renamed in the `foundry-primitives` after removing `AssetAddress`

sgkim126:
@HoOngEe It seems that the snapshot test in dynamic-validator-2 still fails. Would you check it?

test/src/e2e.dynval/1/doubleVote.test.ts
sgkim126: If then, how do you call the 160 bits address that is the hash of the public key? In Foundry, it is address and the platform address is the combination of the network id, version, and address.

HoOngEe: We were calling the `blake160(pub_key)` as `AccountId` in the sdk level. However, it is possible for one to be confused with the Foundry inner type `Address` and the higher level `Address`.

HoOngEe:
@sgkim126 https://github.com/CodeChain-io/foundry/pull/154#issuecomment-595637945. I run the test in the master branch again, but still fails. I think we should fix the test in another pull request. Or am I missing something?

sgkim126:
@HoOngEe The test successes on the master branch. The fragile test is one of the dyn-val-1.

sgkim126:
And I re-ran the Travis several times at the weekend. It seems that the snapshot test fails completely. I suspect that the snapshot test uses ECDSA, but I didn't check it yet.

test/src/e2e.dynval/1/doubleVote.test.ts
sgkim126: Okay. Thank you for explanation.

HoOngEe:
The failure caused by `0x` prefix of `blockHash` and `stateRoot`

HoOngEe:
The root cause of the test failure is the difference between the types in `foundry-rpc-js` and the previous `codechain-sdk`. The hashes in the `block` of `foundry-rpc-js` have `string` types and they have `H256` types in the `sdk` (now they are moved into the test/src).
If you see [the code](https://github.com/CodeChain-io/foundry/blob/ef5d5664eea9125baf83b8d8f776914e0b281f2c/test/src/e2e.dynval/2/snapshot.test.ts#L82), each hash and stateRoot calls `toString()` method. Previously, `H256`'s `toString` method automatically remove `0x`prefixes but now the string instances with `0x` prefixes could not match with the actual snapshot path. Also, `self_nomination_enable` option should also be specified after the auto_self_nomination is merged into `foundry`. I'll add a fix commit

test/tendermint.dynval/snapshot-config.yml
```diff
@@ -1,6 +1,7 @@
 [codechain]
 
 [mining]
+self_nomination_enable = false
```
sgkim126: Why did you change it?

HoOngEe: That's the quick fix for the error 
```
Error: "Error while parsing TOML: Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: \"missing field `self_nomination_enable`\", key: [\"mining\"] } }"
```
We also can designate default value for the field.

PR #162, opened 2020-02-12
Remove the logic related to the score

somniumism:
I removed the `score` in headers. This PR may collide with #153 . To prevent a collision, I removed only the logic related to the `score`, not the field. Structs, which are related to the `score`, still have the `score`. This means that the field not be modified. However, logics such as control flow and some functions were removed. To completely remove the `score`, PR #153  must be merged first. That's on the next issue, and I think it's okay for this PR to merge.

- [x] Remove the logic related to the score in the verification queue
- [x] Remove the logic related to the score in the client
- [x] Remove the logic related to the score in the tendermint
- [x] Remove the logic related to the score in the blockchain
- [x] To check: Replace the value of all scores with U256::default() to check the result
- [x] Remove the score except it of the Header struct and the Block struct

There is no problem when I replaced the value of all scores with a default value.

#### What remains to be done on another issue.
- Remove the score in all the field
- To check: Check e2e tests
- Modify e2e tests for the Foundry without the score


core/src/blockchain/blockchain.rs
```diff
@@ -218,7 +218,7 @@ impl BlockChain {
         let grandparent_hash_of_new_block = parent_details_of_new_block.parent;
         let prev_best_hash = self.best_block_hash();
 
-        if parent_details_of_new_block.total_score + new_header.score() > self.best_proposal_block_detail().total_score
+        if parent_details_of_new_block.number + new_header.number() > self.best_proposal_block_detail().number
```
HoOngEe: This change is weird because the `parent_details_of_new_block.number + new_header.number() > self.best_proposal_block_detail().number` is not the necessary condition of the best block change. I think you can remove the check.

core/src/verification/queue/kind.rs
```diff
@@ -24,6 +23,7 @@ pub use self::headers::Headers;
 use crate::consensus::CodeChainEngine;
 use crate::error::Error;
 use crate::service::ClientIoMessage;
+use primitives::U256;
```
HoOngEe: Please remove the blank line after `use rlp::*;` following the CodeChain-io/codechain#1940.

HoOngEe:
You can also remove score-related errors

sgkim126:
You should consider that the score in tendermint is a combination of a block number and a view count. You cannot just replace a score to a block number. When the block number is the same, you should compare the view of the seal.

core/src/blockchain/blockchain.rs
somniumism: @sgkim126 [What you said](https://github.com/CodeChain-io/foundry/pull/162#issuecomment-586531723) seems to be the case here. @HoOngEe says that this condition is unnecessary, so should I add a condition which is a comparison of `view`? Or should I get rid of all the conditions?

somniumism: To add a condition(comparison of `view`), It seems I have to add `view` to `BlockDetails` and implement calculating the `view` from `seal`.

HoOngEe: @somniumism I just naively thought the clause is removable because it will not affect the behavior. As @sgkim126 said, the logic implies checking the validity of the height and view by the score, the struct may need `view` field.

HoOngEe: @somniumism The below change is also the case. Let's keep the logic equivalent.

somniumism:
@HoOngEe @sgkim126 I added a condition comparing the `view`, but it causes many test failures. Actually I don't know a good way to compare `views`, and I'm not sure what I implemented is right. I need your advice on the design of the condition that compares `views`.

By the way, here, the `score` is defined as follows: `u256_from_u128(std::u128::MAX) * height - view`. And it is used in control flow to determine the best block. And instead of removing the `score`, we should add conditions that compare `numbers` and `views`. So why don't you modify the `score` instead of adding a new condition?

core/src/blockchain/blockchain.rs
```diff
@@ -218,7 +218,12 @@ impl BlockChain {
         let grandparent_hash_of_new_block = parent_details_of_new_block.parent;
         let prev_best_hash = self.best_block_hash();
 
-        if parent_details_of_new_block.total_score + new_header.score() > self.best_proposal_block_detail().total_score
+        let hash = self.best_proposal_block_hash();
+        let header = new_block.header();
+        let parent_header = self.block_header(&parent_hash_of_new_block).expect("Invalid parent hash");
+        let best_header = self.block_header(&hash).expect("Invalid best hash");
+        if parent_header.number() + header.number() > self.best_proposal_block_detail().number
+            && parent_header.view() + header.view() < best_header.view()
```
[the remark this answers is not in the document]
HoOngEe: That's because the direct comparison of `number` and `view` sums are not equivalent to the previous score comparison. Unlike the score comparison, we should not compare the sum of `parent_header` and `header`.  The "if clause" checks if the new block can be a best block. For the condition, I thought the equivalent comparison is `(new_header.number(), self.best_proposal_block_detail().view()) > (self.best_proposal_block_detail().number(), new_header.view())`.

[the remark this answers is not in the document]
somniumism: Thanks for your advice, but could you check my implementation again? Still, 32 tests fail. There seems to be a problem in defining the function `.view()`, but I don't know what the problem is.

HoOngEe:
The failure is caused by `solo` consensus. It does not contain `cur_view` field in the right position as in `Tendermint`, so the general view check is impossible here. We should find another way.

sgkim126:
As @HoOngEe said, you cannot generally reveal view in the core module because of the solo consensus. How about implementing a compare function for the Seal structure?

somniumism:
I added a compare function, and I'm trying to solve 2 test failures in Tendermint:
`Error: Timeout of 90000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/somniumism/Project/PR/foundry/test/src/e2e.long/tendermint.test.ts)`
`Error: Timeout period is expired while waiting best blocknumber` 
I'm looking for the cause.

core/src/views/header.rs
```diff
@@ -112,4 +112,15 @@ impl<'a> HeaderView<'a> {
         let seal = self.seal();
         seal.into_iter().map(|s| rlp::Rlp::new(&s).data().map(|x| x.to_vec())).collect()
     }
+
+    pub fn view(&self) -> Option<u64> {
+        let seal = self.seal();
+        let rlp_view = seal.get(1);
```
HoOngEe: You can utilize the `?` operator for `Option` type.
let rlp_view = seal.get(1)?; then `rlp_view` don't need to be manually unwrapped.

somniumism:
I checked that the test failures occur in the part of comparing views.

core/src/blockchain/headerchain.rs
```diff
@@ -267,20 +272,35 @@ impl HeaderChain {
...
+        let best_proposal_block_detail = self.block_details(&best_proposal_block_hash).expect("Best proposal block always exists");
+
+        let compare_number = new_header.number() > best_proposal_block_detail.number;
+        let compare_view = if new_header.view() == None {
+            true
+        } else {
+            new_header.view() > best_proposal_block_detail.view
+        };
+        
+        compare_number && compare_view
```
HoOngEe: AFAIK, the previous score was designed to pick one block ordered by score so the score formula for Tendermint could be seen arbitrary. However, this number and view compare logic changes the previous ordering.
As I mentioned before, (new_header.number(), best_proposal_block_detail.view()) > (best_proposal_block_detail.number, new_header.view()) seems appropriate because the it follows the previous implication "Pick the higher block number one and pick the lower number one only when the block numbers are equal" but this comparing implies "Pick the one who has higher block number and view".
It seems the test failures are not related directly to this comparing but It's just a comment on consistency.

somniumism:
@HoOngEe I resolved some problems and checked the e2e tests. I'm waiting for your final review. 
By the way, I accidentally chose to merge while resolving the conflict. I'm sorry. Could you resolve it?

core/src/verification/queue/kind.rs
```diff
@@ -14,11 +14,10 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use ctypes::{BlockHash};
-use rlp::*;
-
 pub use self::blocks::Blocks;
 pub use self::headers::Headers;
+use ctypes::BlockHash;
+use rlp::*;
```
HoOngEe: Could you remove the next empty line and reorder the used imports?

core/src/verification/queue/kind.rs
```diff
@@ -14,8 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use ctypes::BlockHash;
-use primitives::U256;
+use ctypes::{BlockHash};
```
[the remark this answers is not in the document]
HoOngEe: remove the empty line for this also.

core/src/blockchain/blockchain.rs
```diff
@@ -210,6 +210,13 @@ impl BlockChain {
         }
     }
 
+    /// Compare the number and the view of current block with these of best block
+    fn compare_number_and_view_to_best(&self, new_header: &HeaderView<'_>) -> bool {
```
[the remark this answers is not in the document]
somniumism: Thank you! I'll modify it to `is_new_header_eligible_to_be_best`

core/src/block.rs
```diff
@@ -208,7 +208,7 @@ impl<'x> OpenBlock<'x> {
 
     /// Populate self from a header.
     fn populate_from(&mut self, header: &Header) {
-        self.block.header.set_score(*header.score());
+        self.block.header.set_score(U256::default());
```
sgkim126: You can write it as `self.block.header.set_score(Default::default());` when the compiler can infer the type.
It's just a comment for further work, you don't need to fix it.

core/src/views/header.rs
```diff
@@ -112,4 +112,12 @@ impl<'a> HeaderView<'a> {
         let seal = self.seal();
         seal.into_iter().map(|s| rlp::Rlp::new(&s).data().map(|x| x.to_vec())).collect()
     }
+
+    /// Get view in the seal field of the header.
+    pub fn view(&self) -> Option<u64> {
```
sgkim126: How about making the view of the solo consensus be always zero instead of returning `None`?

core/src/verification/queue/mod.rs
```diff
@@ -64,10 +64,9 @@ impl Default for Config {
 pub struct VerificationQueue<K: Kind> {
     engine: Arc<dyn CodeChainEngine>,
     verification: Arc<Verification<K>>,
-    processing: RwLock<HashMap<BlockHash, U256>>, // hash to score
+    processing: RwLock<HashMap<BlockHash, U256>>, // hash to block number
```
sgkim126: It seems that we don't use the block number.
How about changing it to HashSet of the block hashes?

somniumism: Like this?
```rust
processing: RwLock<HashMap<BlockHash, U256>>, // HashSet of the block hashes
```

core/src/views/header.rs
somniumism: I implemented to return `None` because the view value is initialized to zero for every new height. I wanted to distinguish the solo consensus from the other consensus. And I implemented the compare function like this:
```rust
    /// Compare the number and the view of current block with these of best block
     fn is_new_header_eligible_to_be_best(&self, new_header: &HeaderView<'_>) -> bool {
         let details_of_best_block = self.best_proposal_block_detail();

          (new_header.number(), details_of_best_block.view) > (details_of_best_block.number, new_header.view())
     }
```
If the view of the solo consensus be always zero instead of returening `None`, It seems to add a new condition which is can tell whether it is zero or not, and the return type needs to be changed from `Option<u64>` to `u64`. How about this?

core/src/verification/queue/kind.rs
```diff
@@ -104,8 +103,8 @@ pub mod headers {
             *self.parent_hash()
         }
 
-        fn score(&self) -> U256 {
-            *self.score()
+        fn block_number(&self) -> U256 {
```
HoOngEe: BTW, we are using the `BlockNumber` type for block numbers which is an alias type of `u64`. Do we need to use the looser type `U256` ?

core/src/verification/queue/mod.rs
HoOngEe: like this processing: RwLock<HashSet<BlockHash>>

somniumism: I see. Thank you : )

core/src/views/header.rs
sgkim126: There is no difference between returning None and retuning 0.
In your implementation, the view of the solo is always None, so the `is_new_header_eligible_to_be_best` function just compares the block number.

core/src/verification/queue/kind.rs
somniumism: These functions seem unnecessary and are unused, so I removed these functions. Thank you : )

core/src/verification/queue/mod.rs
somniumism: I modified it! : )

core/src/views/header.rs
somniumism: I modified it. Thank you : )

somniumism:
@HoOngEe @sgkim126 All requested changes were modified. Could you check this PR again? By the way, This PR consists of 4 commits. Is it okay to be merged without having to squash?

core/src/verification/queue/kind.rs
HoOngEe: I think you should remove the function from the original commit not remove the logic in the last commit.

core/src/blockchain/extras.rs
```diff
@@ -93,10 +93,10 @@ impl Key<TransactionAddresses> for Tracker {
 /// Familial details concerning a block
 #[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
 pub struct BlockDetails {
+    /// Block view
+    pub view: u64,
```
HoOngEe: Though It is just a type alias but what about using `View` type defined in `core/types`

core/src/verification/queue/kind.rs
somniumism: You mean remove the function from the commit `Remove the logic related to the score in queue` commit?

HoOngEe: Yes, unnecessary changes don't have to exist in commit histories.

somniumism: I modified it : ) Thank you for your good advice!

sgkim126:
> Is it okay to be merged without having to squash?

I would merge the commits into one, but if it can be compiled for each commit, I don't matter even they are divided.

core/src/client/test_client.rs
```diff
@@ -136,7 +136,7 @@ impl TestBlockChainClient {
         let genesis_block = scheme.genesis_block();
         let genesis_header = scheme.genesis_header();
         let genesis_hash = genesis_header.hash();
-        let genesis_score = *genesis_header.score();
+        let genesis_score = U256::default();
```
sgkim126: Who uses it? Why the score remains after this pr?

sgkim126:
Why does this PR remove the logic related to score, but doesn't remove the score field?

core/src/client/test_client.rs
somniumism: I didn't completely remove the score from this PR. I discussed this work with @majecty  and decided not to remove the score from the Header struct and the Block struct and the fields related to these. It is to prevent conflicts with other team members currently using or modifying this part. According to the @majecty , changing the header and block field, the logic and e2e test in relation to it can cause a collision, so we don't have to do this right now. So I didn't change some fields, only removed the score that had nothing to do with the logic involved and removal.

majecty:
@sgkim126 I requested @somniumism not to remove the score field.

There were PRs that will conflict with this PR. One PR was adding a field in the header struct. The other PR was changing the whole tests.

I thought we can split removing the score into two PRs. One is not using the score. The other one is removing the score field.

IMO, removing the score field after all the big changes in the test directory will be easier.

core/src/client/test_client.rs
somniumism: By the way, this logic `let genesis_score = *genesis_header.score()` seems unnecessary. It's my mistake.

somniumism: @sgkim126 I modified it, and all the unnecessary parts that were not removed were also removed. Thank you very much for your careful review. In that sense, could you check it one more time? : )))

somniumism:
> I would merge the commits into one, but if it can be compiled for each commit, I don't matter even they are divided.

Each commit is normally built. When the review is finished, I will merge the four commits.

PR #166, opened 2020-02-13
Implement connOpenTry datagram in the ICS connection spec

majecty:


core/src/ibc/connection_03/manager.rs
```diff
@@ -24,6 +24,11 @@ use rlp::{Encodable, Rlp};
 #[derive(Default)]
 pub struct Manager {}
```
junha1: no context?

core/src/ibc/transaction_handler/datagrams.rs
```diff
@@ -93,6 +106,29 @@ impl Encodable for Datagram {
...
+                counterparty_prefix,
+                counterparty_client_identifier,
+                client_identifier,
+                proof_init,
+                proof_consensus,
+                proof_height,
+                consensus_height,
+            } => {
+                s.begin_list(10);
+                s.append(&DatagramTag::ConnOpenTry)
```
junha1: I think we really need to improve macro to support `enum` and `tuple` someday...

core/src/ibc/connection_03/manager.rs
```diff
@@ -55,6 +60,105 @@ impl Manager {
...
+        if kv_store.has(&path) {
+            let raw = kv_store.get(&path);
+            let connection_end = rlp::decode(&raw).expect("Only the connection code can save the code");
+            return Some(connection_end)
+        }
+
+        None
+    }
+
+    fn verify_connection_state(
```
junha1: This will be moved to somewhere in `/client_02/`?

core/src/ibc/connection_03/manager.rs
majecty: Let' add context at the same time in another PR. I was creating this PR before the discussion.

core/src/ibc/connection_03/manager.rs
majecty: Yes, the code will be moved after we update the client_02 code.

core/src/ibc/connection_03/manager.rs
junha1: Ok no problem

core/src/ibc/connection_03/manager.rs
```diff
@@ -55,6 +60,105 @@ impl Manager {
...
+        &self,
+        ctx: &mut dyn ibc::Context,
+        connection: &ConnectionEnd,
```
junha1: Could you explain the meaning of this argument?

majecty: I checked the spec. The spec is not using it. Please ignore the argument.

PR #170, opened 2020-02-14
Add get_current_height function to the IBC Context

majecty:
The connection spec's OpenTry datagram handler need to know the host's current height.
https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics

### DIFF
--- a/core/src/client/client.rs
+++ b/core/src/client/client.rs
@@ -867,7 +867,8 @@ impl IBCTransactionExecutor for Client {
         state: &mut TopLevelState,
         fee_payer: &Address,
         sender_pubkey: &Public,
+        current_block_number: BlockNumber,
     ) -> StateResult<()> {
-        ibc::execute_transaction(bytes, state, fee_payer, sender_pubkey)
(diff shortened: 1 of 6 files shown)


junha1:
current height of mine or other's?

majecty:
The current height of host.
You can find the "getCurrentHeight" definition in this spec document: https://github.com/cosmos/ics/tree/master/spec/ics-024-host-requirements#consensus-state-introspection

junha1:
Ah ok. It makes sense

junha1:
Btw, have you considered that latest height is always available via `ClientState`?

majecty:
@junha1 The height in the ClientState is not the host's height. The height is for the counterparty chain.

> ClientState is an opaque data structure that keeps arbitrary internal state to track verified roots and past misbehaviours. In our case, the hash of next validator set + latest height + frozen height(optional)
https://research.codechain.io/t/foundry-light-client/41

junha1:
@majecty Ah, my mistake.

PR #175, opened 2020-02-15
Make the foundry use the dynamic validators from the genesis block

sgkim126:


core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -25,64 +25,32 @@ use ctypes::BlockHash;
...
 
 impl DynamicValidator {
-    pub fn new(initial_validators: Vec<Public>) -> Self {
-        DynamicValidator {
-            initial_list: RoundRobinValidator::new(initial_validators),
-            client: Default::default(),
-        }
-    }
-
     fn next_validators(&self, hash: BlockHash) -> Option<Vec<Validator>> {
```
majecty: We can remove Option here.

core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -25,64 +25,32 @@ use ctypes::BlockHash;
...
-            let mut validators: Vec<_> = validators.into();
-            validators.reverse();
-            Some(validators)
-        }
+        let mut validators: Vec<_> = validators.into();
+        validators.reverse();
+        Some(validators)
     }
 
     fn current_validators(&self, hash: BlockHash) -> Option<Vec<Validator>> {
```
majecty: ditto.

core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -99,17 +67,14 @@ impl DynamicValidator {
         }
```
majecty: Shouldn't we remove this?

core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -137,162 +98,123 @@ impl DynamicValidator {
...
-        if let Some(validators) = self.validators_pubkey(*parent) {
-            let n_validators = validators.len();
-            *validators.get(index % n_validators).unwrap()
-        } else {
-            self.initial_list.get(parent, index)
+    fn get(&self, parent: &BlockHash, index: usize) -> Option<Public> {
+        let validators = self.validators_pubkey(*parent)?;
+        if validators.is_empty() {
+            return None
         }
```
majecty: validators will not be empty.

core/src/consensus/stake/mod.rs
```diff
@@ -68,8 +77,43 @@ impl ActionHandler for Stake {
...
+                        RuntimeError::FailedToHandleCustomAction("Can delegate to who is a candidate".into()).into()
+                    )
+                }
+            }
+            *stake -= total_delegation;
+        }
+
+        for (address, deposit) in &self.genesis_candidates {
+            let stake = genesis_stakes.entry(*address).or_default();
+            if *stake < deposit.deposit {
```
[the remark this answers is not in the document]
sgkim126: I confused. It's my mistake. I'll fix it.

core/src/consensus/validator_set/dynamic_validator.rs
sgkim126: You are right. I'll change it.

core/src/consensus/validator_set/dynamic_validator.rs
sgkim126: Yes, we should remove it.
I'll do it and remove PreviousValidators stored in the state in another PR.

sgkim126: I added the commit removing the code related to the previous validators.
I planed to make the other PR because actually it should be done by https://github.com/CodeChain-io/foundry/pull/277, but I think adding the commit in this PR also seems good.

core/src/consensus/stake/mod.rs
```diff
@@ -78,6 +107,45 @@ impl ActionHandler for Stake {
...
+                cerror!(STATE, "{} has insufficient balance to become the candidate", address);
+                return Err(RuntimeError::InsufficientBalance {
+                    address: *address,
+                    balance,
+                    cost: deposit.deposit,
+                }
+                .into())
+            }
+            state.sub_balance(address, deposit.deposit).unwrap();
+        }
```
majecty: It's perfect!

PR #180, opened 2020-02-17
Implement ConnOpenAck datagrain in the ICS Connection spec

majecty:


core/src/ibc/connection_03/manager.rs
```diff
@@ -122,6 +122,45 @@ impl<'a> Manager<'a> {
...
+            return Err(format!("Invalid connection state expected INIT or TRYOPEN but found {:?}", connection.state))
+        }
+        let expected_connection = ConnectionEnd {
+            state: ConnectionState::TRYOPEN,
+            counterparty_connection_identifier: identifier.clone(),
+            counterparty_prefix: get_commiment_prefix(),
+            client_identifier: connection.counterparty_client_identifier.clone(),
+            counterparty_client_identifier: connection.client_identifier.clone(),
+        };
+        self.verify_connection_state(&connection, proof_height, proof_try, identifier.clone(), &expected_connection);
```
junha1: If this function will be moved to client, how will you retrieve a reference to client?

majecty: `ConnectionEnd` contains the client identifier.

junha1: We concluded that all `verify_...` function will take ID of client.

majecty:
@junha1 I added this commit: https://github.com/CodeChain-io/foundry/pull/180/commits/52cc5fd7a8bd239b7b82ea2b8b5810534a7cc9e8

Other things are barely changed. I changed the type of proof to `Vec<u8>` from `String`. I changed the code that was using `get_kv_store` to use `get_kv_store_mut`.

core/src/ibc/connection_03/manager.rs
```diff
@@ -135,13 +174,19 @@ impl<'a> Manager<'a> {
         &mut self,
         connection: &ConnectionEnd,
         proof_height: u64,
-        proof: CommitmentProof,
+        proof: Vec<u8>,
         connection_identifier: Identifier,
         connection_end: &ConnectionEnd,
     ) -> bool {
+        let proof_dec: CommitmentProof = if let Ok(proof) = rlp::decode(&proof) {
```
junha1: decoding should happen only in the client.

majecty: I think that the whole `verify_connection_state` function will be moved to the light client soon.

junha1: Ah ok. You just wanted to keep that consistent. I'll move it soon, so never mind.

PR #182, opened 2020-02-18
Refactor transaction execution logic

### LINKED ISSUE #160: Refactor transaction execution
After #155, we can assert that the address of an account is always the hash of its public key.
Therefore, transaction execution logic can be simplified.

Byeongjee:
It closes #160 

It was hard to understand the context with the previous parameter name `fee_payer`.
It also reduces cohesion between `Address` and `Public`, which will allow easier landing of #151


majecty:
@sgkim126 @HoOngEe I added both of you as reviewers to notify you that the name of the fee_payer field is changed.
You do not need to review whole changes.
I'll review them.

majecty:
@Byeongjee Please do not merge a PR until all reviewers click the "approve" button.

Byeongjee:
Sorry. I got it.

PR #184, opened 2020-02-18
Modify test files based on local sdk and stakeholder

MSNTCS:


test/src/stakeholder/src/rpc.ts
```diff
@@ -1,5 +1,6 @@
-import { SDK } from "codechain-sdk";
-import { PlatformAddress } from "codechain-sdk/lib/core/classes";
+// @ts-ignore
```
majecty: Why did you add this line?

MSNTCS:
@majecty,  the failed test was run perfectly in my machine and also was run without any problem here https://travis-ci.com/MSNTCS/foundry.

PR #185, opened 2020-02-18
Implement ConnOpenConfirm datagram

majecty:
~Depends on https://github.com/CodeChain-io/foundry/pull/180~

### DIFF
--- a/core/src/ibc/transaction_handler/datagrams.rs
+++ b/core/src/ibc/transaction_handler/datagrams.rs
@@ -251,3 +251,49 @@ impl Decodable for Datagram {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rlp::{self, rlp_encode_and_decode_test};
+
+    #[test]
+    fn conn_open_init() {
+        let conn_open_init = Datagram::ConnOpenInit {
+            identifier: "identifier".to_owned(),
+            desired_counterparty_connection_identifier: "desired_counterparty_connection_identifier".to_owned(),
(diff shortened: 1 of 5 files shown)


core/src/ibc/transaction_handler/datagrams.rs
```diff
@@ -83,6 +85,11 @@ pub enum Datagram {
         proof_height: u64,
         consensus_height: u64,
     },
+    ConnOpenConfirm {
+        identifier: String,
+        proof_ack: Vec<u8>,
```
junha1: How about `Bytes` for proofs and `ctypes::BlockNumber` for heights? I'll update previous ones (including this).

majecty: I'll create a new commit that changes Vec<u8> to Bytes in all IBC code.
IMO, we should not use the ctypes::BlockNumber here. The should consider other chain's block number. `ctypes::BlockNumber` is for Foundry only.

junha1: Ok then I'll leave `Vec<u8>`s. And please contain update to `ctypes::BlockNumber` for *our blocks*. (e.g *making* proof)

majecty: How about defining a `IBCBlockNumber` type instead?
I'm not sure which method is better, but using a `IBCBlockNumber` type in the all IBC code is easier.

junha1: Maybe together with `Identifier`

PR #186, opened 2020-02-19
Refine sdk and stakeholder directories 

MSNTCS:
Currently, due to the fact that we merge `codechain-sdk` and `codechian-stakeholder`, there were a lot of unuseful files. In this patch, we remove all unnecessary files in both sdk and stakeholder directories.


test/package.json
```diff
@@ -31,6 +31,7 @@
     "@types/lodash": "^4.14.149",
     "@types/mkdirp": "^0.5.2",
     "@types/mocha": "^5.2.7",
+    "@types/bn.js": "^4.11.5",
```
majecty: Can you explain why did you add this change in this commit?

test/package.json
```diff
@@ -25,38 +25,50 @@
     "fmt": "tslint -p . --fix && prettier 'src/**/*.{ts, json}' --write"
   },
   "devDependencies": {
+    "@types/jest": "^23.3.1",
```
majecty: Please add a description of why you added these dependencies in the commit.

.travis.yml
```diff
@@ -23,9 +23,6 @@ jobs:
         - cargo fetch --verbose
         - cargo build
         - cd test && yarn
-        - cd src/sdk/ && yarn
```
majecty: Could you add a brief description of why we added these lines before and removed them again?

test/package.json
majecty: Please remove thins line from "Remove ccstake from stakeholder " commit.

PR #172, opened 2020-02-14
Implement client_02 and commitment_23

junha1:
Closes #112
Closes #126

PR #188, opened 2020-02-19
Refactor CompactValidatorSet to impl Deref

junha1:
As @majecty suggested https://github.com/CodeChain-io/foundry/pull/172#discussion_r381087816


junha1:
Failed test is time-out I think.

PR #195, opened 2020-02-21
Impement queryConnection and queryClientConnections

majecty:
~I left the proof part. I'll update this PR after Junha finish https://github.com/CodeChain-io/foundry/pull/196~
Now, I'm ready for review.


core/src/ibc/connection_03/manager.rs
```diff
@@ -221,6 +221,12 @@ impl<'a> Manager<'a> {
         Some(connection_end)
     }
 
+    pub fn query_client_connections(&self, identifier: &str) -> ConnectionIdentifiersInClient {
```
junha1: Let's make all `Manager`'s query functions to be private.

majecty: I fixed them.

PR #196, opened 2020-02-21
Implement RPCs for client_02

### LINKED ISSUE #192: Implement basic RPCs for ICS
- `QueryClientState(id) -> (value, height, proof)` : `height` may be omitted if the response is given in `json`
- `QueryConsensusState(id, height) -> (value, proof)`
- `ComposeHeader(id, height) -> header` : `header` is ok to be bytes. (Relayer doesn't care about the contents)

junha1:
#Closes #192

### DIFF
--- a/core/src/ibc/querier/mod.rs
+++ b/core/src/ibc/querier/mod.rs
@@ -1,0 +1,96 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// This module representes all accessible interface
+use crate::ibc;
+use ibc::commitment_23::types::{create_membership_proof, create_non_membership_proof};
+use ibc::commitment_23::{CommitmentPath, CommitmentState};
+use primitives::Bytes;
+use rlp::Decodable;
+
+use crate::ibc::IdentifierSlice;
+
+pub trait DebugName {
+    fn debug_name() -> &'static str;
+}
+
+impl DebugName for ibc::connection_03::types::ConnectionEnd {
+    fn debug_name() -> &'static str {
+        "ConnectionEnd"
+    }
+}
+
+impl DebugName for ibc::client_02::types::ClientState {
+    fn debug_name() -> &'static str {
+        "ClientState"
+    }
+}
+
+impl DebugName for ibc::client_02::types::ConsensusState {
+    fn debug_name() -> &'static str {
+        "ConsensusState"
+    }
+}
+
+/// Queries the path and returns the result in decoded struct
+pub fn query<T>(ctx: &dyn ibc::Context, path: &CommitmentPath) -> Option<T>
+where
+    T: Decodable + DebugName, {
+    if ctx.get_kv_store().has(&path.raw) {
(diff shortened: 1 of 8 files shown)


junha1:
@majecty Please take a look at the interface. In short, `compose_header()` returns opaque(`Bytes`) data, whereas `query_...()` return transparent(`JSON`) data.

majecty:
@junha1 The interface looks good to me.

core/src/ibc/querier/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+/// This module representes all accessible interface
+use crate::ibc;
+use ibc::commitment_23::types::{create_membership_proof, create_non_membership_proof};
+use ibc::commitment_23::{CommitmentPath, CommitmentState};
+use primitives::Bytes;
+use rlp::Decodable;
+
+use crate::ibc::IdentifierSlice;
+
+pub trait Nameable {
```
majecty: The `Nameable` is used to print a debug message.
How about changing it like `DebugName`?
Then the `name()` should be changed to `debug_name`.

I found a discussion about trait name and it looks good to me.
> If the trait has a single self-explanatory method (or a set of nearly identical methods), name it after the method: Clone, Hash, Default, Into, Write, ToOwned, AsRef, Extend.

https://github.com/rust-lang/api-guidelines/issues/28#issuecomment-418509628

rpc/src/v1/types/ibc.rs
```diff
@@ -13,3 +13,62 @@
...
+/// Because of the nature of IBC, they commonly
+/// 1. Requires a block number for which proof stands
+/// 2. The data should be transparent:
+/// relayer must be able to open it and extract required infomation
+/// 3. Includes a cryptographical proof of that
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IBCQuery {
+    pub number: u64,
+    pub data: String,
```
majecty: Shouldn't we use the data as a generic parameter T?

rpc/src/v1/types/ibc.rs
```diff
@@ -13,3 +13,62 @@
...
+    pub proof: Bytes,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IBCQueryAbsence {
+    pub number: u64,
+    pub proof: Bytes,
+}
```
majecty: When we will use `IBCQueryAbsence` type?
IMO, we should make `IBCQuery.data` as optionable.

junha1:
@majecty I updated all

rpc/src/v1/types/ibc.rs
```diff
@@ -13,3 +13,57 @@
...
+/// Note : proof may represents both that of presence and absence. It depends on option of data.
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IBCQuery<T: Serialize> {
+    pub number: u64,
+    pub data: Option<T>,
+    pub proof: Bytes,
+}
+
+/* -------- Client 02 -------- */
```
majecty: We are not using `/*` comment.
Please use `///` and use a full sentence.

rpc/src/v1/types/ibc.rs
```diff
@@ -13,3 +13,57 @@
...
+    /// Unpacked light_client::ClientState
+    pub number: u64,
+    pub next_validator_set_hash: H256,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ConsensusState {
+    pub validator_set_hash: H256,
+    // Unpacked CommitmentRoot
```
majecty: Please us `///`

rpc/src/v1/types/ibc.rs
```diff
@@ -13,3 +13,58 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct IBCQuery<T: Serialize> {
+    pub number: u64,
+    pub data: Option<T>,
+    pub proof: Bytes,
+}
+
+/// Client 02 related types
+
```
junha1: I added an extra line to say that this comment is not only for the one right after  (`ClientState`).

PR #198, opened 2020-02-23
Cleanup code and upgrade rust toolchains

sgkim126:


Cargo.toml
```diff
@@ -17,17 +17,17 @@ edition = "2018"
...
+ccore = { package = "codechain-core", path = "core" }
+cdiscovery = { package = "codechain-discovery", path = "discovery" }
+codechain-logger = { path = "util/logger" }
+ckey = { package = "codechain-key", path = "key" }
+ctypes = { package = "codechain-types", path = "types" }
+ckeystore = { package = "codechain-keystore", path = "keystore" }
+cnetwork = { package = "codechain-network", path = "network" }
+crpc = { package = "codechain-rpc", path = "rpc" }
+csync = { package = "codechain-sync", path = "sync" }
+ctimer = { package = "codechain-timer", path = "util/timer" }
```
majecty: I'm happy to see these changes!
Great work!

majecty:
@sgkim126 Many tests failed. Do you know the reason?

sgkim126:
@majecty I'll check it.

sgkim126:
It's because a lock held too long.
https://github.com/CodeChain-io/foundry/pull/198/commits/e42015fcb6d4257b979d193fd6b2dee4faee6025#diff-2e6fb877a7bd787856f23dae2d597480R154
I changed the variable name because the clippy warns not to use _ for a lock name.
As you know, because the underscore doesn't hold the variable, the lock is released right after the statement in the current code.
The clippy warns that I should change the name or explicitly drop it, so I changed the name. It changed the life-time of the lock.
I thought it's okay to hold the lock for the function, but it should be released before joining the worker thread because the worker thread also acquires the lock.

PR #203, opened 2020-02-24
Remove two empty lines

majecty:
We are using only one empty line to split code.


sgkim126:
@majecty It seems that the issue title does not correspond to the commit.

majecty:
@sgkim126 Sorry I fixed it.

Issue #144, opened 2020-02-07
Remove unused testnet settings


kseo:
Remove `core/res/beagle.json` and `core/res/corgi.json`.

PR #209, opened 2020-02-27
Remove unnecessary chains

sgkim126:
Because they are chains for CodeChain, Foundry will need new test chains and mainnet.

It closes https://github.com/CodeChain-io/foundry/issues/144.


junha1:
https://github.com/CodeChain-io/foundry/issues/144

PR #210, opened 2020-02-27
Implement basics of ICS 04

junha1:


core/src/ibc/channel_04/manager.rs
```diff
@@ -40,4 +44,61 @@ impl<'a> Manager<'a> {
...
+        &mut self,
+        order: ChannelOrder,
+        connection: Identifier,
+        channel_identifier: Identifier,
+        counterparty_channel_identifier: Identifier,
+        version: String,
+    ) -> Result<Identifier, String> {
+        let kv_store = self.ctx.get_kv_store_mut();
+
+        // It is ok to be in any state.
```
majecty: Could you add a detailed description of this?
"It is ok to be in any state" does not give the intention to the reader.

The ICS spec is saying "optimistic" at the same point.
After reading it, I can understand the behavior. Channel can be initiated while the connection is establishing.

So, when you write a comment, please explain the intention of code, not the behavior of code.

core/src/ibc/channel_04/manager.rs
```diff
@@ -40,4 +44,61 @@ impl<'a> Manager<'a> {
...
+        let channel = ChannelEnd {
+            state: ChannelState::INIT,
+            ordering: order,
+            counterparty_port_identifier: DEFAULT_PORT.to_string(),
+            counterparty_channel_identifier,
+            connection_hops: vec![connection],
+            version,
+        };
+
+        // NOTE: You don't have to recover the state even if the transaction fails.
```
majecty: This behavior is common in all transactions in Foundry. Please do not write it in this place.

core/src/ibc/channel_04/manager.rs
```diff
@@ -40,4 +44,61 @@ impl<'a> Manager<'a> {
...
+        // NOTE2: Is it ok to rewrite ChannelEnd when it exists? Currently it fails the transaction.
```
majecty: Please write concerns in an issue, not in the comment. We can't discuss using comments.

majecty: Please write a result of a discussion in a comment.

core/src/ibc/channel_04/manager.rs
```diff
@@ -101,4 +103,103 @@ impl<'a> Manager<'a> {
...
+        proof_init: Bytes,
+        proof_height: u64,
+    ) -> Result<Identifier, String> {
+        let kv_store = self.ctx.get_kv_store();
+
+        let previous = kv_store.get(&channel_path(&channel_identifier, DEFAULT_PORT));
+
+        if let Some(x) = previous {
+            let channel_end: ChannelEnd = rlp::decode(&x).expect("Illformed connection end stored in the DB");
+            if channel_end.state != ChannelState::INIT
```
majecty: How about creating an expected channel and comparing the expected and `channel_end`?

core/src/ibc/channel_04/manager.rs
```diff
@@ -101,4 +103,103 @@ impl<'a> Manager<'a> {
...
+                || channel_end.ordering != order
+                || channel_end.counterparty_port_identifier != DEFAULT_PORT
+                || channel_end.counterparty_channel_identifier != counterparty_channel_identifier
+                || channel_end.connection_hops != vec![connection.clone()]
+                || channel_end.version != version
+            {
+                return Err("There already exists ChannelEnd on which open_try() can't be conducted.".to_owned())
+            }
+        } // ok to be not found.
```
majecty: Ditto. Comment should describe intention.

majecty: https://www.codeconquest.com/advanced-programming-concepts/code-commenting/

core/src/ibc/channel_04/manager.rs
```diff
@@ -101,4 +103,103 @@ impl<'a> Manager<'a> {
...
+        if connection_end.state != ConnectionState::OPEN {
+            return Err("Connection not opend".to_owned())
+        }
+
+        let expected = ChannelEnd {
+            state: ChannelState::INIT,
+            ordering: order,
+            counterparty_port_identifier: DEFAULT_PORT.to_string(),
+            counterparty_channel_identifier: channel_identifier.clone(),
+            connection_hops: vec![connection.clone()], // Note: the array should be reversed in the future where `connection` becomes an array.
```
majecty: Please write comment in a new line.

core/src/ibc/channel_04/manager.rs
```diff
@@ -101,4 +103,103 @@ impl<'a> Manager<'a> {
...
+            ordering: order,
+            counterparty_port_identifier: DEFAULT_PORT.to_string(),
+            counterparty_channel_identifier,
+            connection_hops: vec![connection],
+            version,
+        };
+
+        let kv_store = self.ctx.get_kv_store_mut();
+
+        // NOTE: You don't have to recover the state even if the transaction fails.
```
majecty: ditto.

core/src/ibc/channel_04/manager.rs
```diff
@@ -101,4 +103,103 @@ impl<'a> Manager<'a> {
...
+        // NOTE2: Is it ok to rewrite ChannelEnd when it exists? Currently it fails the transaction.
```
majecty: ditto.

core/src/ibc/client_02/manager.rs
```diff
@@ -91,34 +91,46 @@ impl<'a> Manager<'a> {
...
+    pub fn verify_connection_state(
+        &self,
+        id: IdentifierSlice,
+        proof_height: BlockNumber,
+        proof: Bytes,
+        connection_identifier: IdentifierSlice,
+        connection_end: &ConnectionEnd,
+    ) -> Result<(), String> {
+        let path = format!("connections/{}", connection_identifier);
+        let value_dec = rlp::encode(connection_end);
```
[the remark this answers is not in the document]
junha1: it is renamed in next commit. This commit purely contains extracting a common logic.

core/src/ibc/channel_04/manager.rs
junha1: Actually that link is basically saying that  `ok to be not found` is appropriate comment here, because I clarified that it was not a mistake to miss an `else` clause.

majecty: How about changing the comment "We can move to the `ChanOpenTry` state from `Uninitialized` state"?

junha1:
@majecty I updated. Please check.

core/src/ibc/channel_04/manager.rs
```diff
@@ -40,4 +44,60 @@ impl<'a> Manager<'a> {
...
+        order: ChannelOrder,
+        connection: Identifier,
+        channel_identifier: Identifier,
+        counterparty_channel_identifier: Identifier,
+        version: String,
+    ) -> Result<Identifier, String> {
+        let kv_store = self.ctx.get_kv_store_mut();
+
+        // It is ok to be in any state, since here we do 'optimistic' handshake, where we establish a channel while the connection is not established completely.
+        // Thus we check only the existence.
```
majecty: Good explanation!

core/src/ibc/channel_04/manager.rs
```diff
@@ -100,4 +102,105 @@ impl<'a> Manager<'a> {
...
+        proof_init: Bytes,
+        proof_height: u64,
+    ) -> Result<Identifier, String> {
+        let kv_store = self.ctx.get_kv_store();
+
+        let previous = kv_store.get(&channel_path(&channel_identifier, DEFAULT_PORT));
+
+        // A trivial case is establishing a channel from scratch, so there won't be an existing one while responding counterparty's 'init' with 'try'.
+        // However, if it exists, we just check that our last trial ('init') was intended to make a same channel as the counterparty chain's 'init' did.
+        // And then overwrites it.
```
majecty: :+1:

PR #214, opened 2020-02-28
Implement initial design of Foundry ABCI

Byeongjee:


Byeongjee:
Some fields, such as `Event` and `Log`, are not used in current Foundry. Should I keep them?

dynaxis:
> Some fields, such as `Event` and `Log`, are not used in current Foundry. Should I keep them?
We definitely need all of them IMO. Since @majecty wants to discuss why ICS needs deterministic log today, let's decide on the details. I think we need to coin appropriate terms to designate both deterministic and non-deterministic logs.

dynaxis:
IMHO, the current design is unnecessarily sticking to the ABCI style. ABCI is implemented over gRPC (thus using protocol buffer), and I guess it's why ABCI has plenty of `Request...`s and `Response...`s. I think our interface should be in natural Rust style. I recommend you to polish up the style of the interface a bit before merging this PR. I'll also try to assist you in the course.

Byeongjee:
> IMHO, the current design is unnecessarily sticking to the ABCI style. ABCI is implemented over gRPC (thus using protocol buffer), and I guess it's why ABCI has plenty of `Request...`s and `Response...`s. I think our interface should be in natural Rust style. I recommend you to polish up the style of the interface a bit before merging this PR. I'll also try to assist you in the course.

I removed `Request` and `Response` in names.
However, I'm not very satisfied with the new names `TransactionCheckResult`, `TransactionExecutionResult`, ... because the suffix `Result` might be confused with the Rust `Result` type. How do you think about it? Could you suggest a better name?

Byeongjee:
Also, I merged the `commit()` method into `execute_block()` method. How do you think of it?

dynaxis:
I'll try to propose changes to the last update today. Anyway, with the last change, it got better. Thanks.

dynaxis:
> Also, I merged the `commit()` method into `execute_block()` method. How do you think of it?

Seems reasonable. `commit` in ABCI was necessary since it delivers Tx multiple times.

Byeongjee:
I removed `Log` from ABCI returns since it was our conclusion from the meeting today.
I changed the return type of `check_transaction`, so there are three options now.
- `IncludeInCurrentBlock`
- `IncludeInFutureBlock`
- `Exclude`
I changed this to make it compatible with current Foundry implementation.

PR #215, opened 2020-02-28
Create a script that runs two Foundry networks to use it in the ics scenario

majecty:


junha1:
This is the only script that will be tested as PoC?

ibc.ts/src/scenario/runChains.ts
```diff
@@ -0,0 +1,175 @@
...
+        ["-c", "./chainB.schem.json", "--config", "./chainB.config.toml"],
+        {
+            cwd: "./chainB"
+        }
+    );
+    streamOutToDebug(foundryProcess);
+}
+
+async function checkChainAAndBAreRunning() {
+    // Wait for Foundry to listen on the port, three seconds is a random value.
```
junha1: a random value -> just an arbitrary value?

majecty: Good. I'll fix it to "Wait for Foundry to listen on the port; three seconds is an arbitrary value."

majecty: @junha1 I fixed it.

PR #219, opened 2020-03-02
Hot events for Informer API 

MSNTCS:
Informer API is supposed to send events as soon as they are triggered to all subscribed clients. In this PR, Hot events, those are not retrieved from the history of the event, are covered.

### DIFF
--- a/foundry/config/mod.rs
+++ b/foundry/config/mod.rs
@@ -18,6 +18,7 @@ mod chain_type;
 
 use ccore::{MemPoolMinFees, MinerOptions, TimeGapParams};
 use cidr::IpCidr;
+use cinformer::InformerConfig;
 use ckey::PlatformAddress;
 use cnetwork::{FilterEntry, NetworkConfig, SocketAddr};
 use primitives::H256;
@@ -38,6 +39,7 @@ pub struct Config {
     pub network: Network,
     pub rpc: Rpc,
     pub ws: Ws,
+    pub informer: Informer,
     pub snapshot: Snapshot,
     #[serde(default)]
     pub email_alarm: EmailAlarm,
@@ -51,6 +53,7 @@ impl Config {
         self.network.merge(&other.network);
         self.rpc.merge(&other.rpc);
         self.ws.merge(&other.ws);
+        self.informer.merge(&other.informer);
         self.snapshot.merge(&other.snapshot);
         self.email_alarm.merge(&other.email_alarm);
     }
@@ -122,6 +125,16 @@ impl Config {
             interface: self.ws.interface.clone().unwrap(),
             port: self.ws.port.unwrap(),
             max_connections: self.ws.max_connections.unwrap(),
+        }
+    }
+
+    pub fn informer_config(&self) -> InformerConfig {
+        debug_assert!(!self.informer.disable.unwrap());
+
+        InformerConfig {
+            interface: self.informer.interface.clone().unwrap(),
+            port: self.informer.port.unwrap(),
+            max_connections: self.informer.max_connections.unwrap(),
         }
     }
 
@@ -270,6 +283,15 @@ pub struct Ws {
 
 fn default_enable_devel_api() -> bool {
     cfg!(debug_assertions)
+}
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Informer {
+    pub disable: Option<bool>,
+    pub interface: Option<String>,
+    pub port: Option<u16>,
+    pub max_connections: Option<usize>,
 }
 
 #[derive(Deserialize)]
@@ -641,6 +663,39 @@ impl Rpc {
     }
 }
 
+impl Informer {
+    pub fn merge(&mut self, other: &Informer) {
+        if other.disable.is_some() {
+            self.disable = other.disable;
+        }
+        if other.interface.is_some() {
+            self.interface = other.interface.clone();
+        }
+        if other.port.is_some() {
+            self.port = other.port;
+        }
+        if other.max_connections.is_some() {
+            self.max_connections = other.max_connections;
+        }
+    }
+    pub fn overwrite_with(&mut self, matches: &clap::ArgMatches<'_>) -> Result<(), String> {
+        if matches.is_present("no-informer") {
+            self.disable = Some(true);
+        }
+
+        if let Some(interface) = matches.value_of("informer-interface") {
+            self.interface = Some(interface.to_string());
+        }
+        if let Some(port) = matches.value_of("informer-port") {
+            self.port = Some(port.parse().map_err(|_| "Invalid port")?);
+        }
+        if let Some(max_connections) = matches.value_of("informer-max-connections") {
+            self.max_connections = Some(max_connections.parse().map_err(|_| "Invalid max connections")?);
+        }
+        Ok(())
+    }
+}
+
 impl Ws {
     pub fn merge(&mut self, other: &Ws) {
         if other.disable.is_some() {
@@ -769,6 +824,7 @@ pub fn load_config(matches: &clap::ArgMatches<'_>) -> Result<Config, String> {
     config.network.overwrite_with(&matches)?;
     config.rpc.overwrite_with(&matches)?;
     config.ws.overwrite_with(&matches)?;
+    config.informer.overwrite_with(&matches)?;
     config.snapshot.overwrite_with(&matches)?;
     config.email_alarm.overwrite_with(&matches)?;
     Ok(config)
--- a/foundry/run_node.rs
+++ b/foundry/run_node.rs
@@ -27,12 +27,15 @@ use ccore::{
     MinerService, PeerDb, Scheme, NUM_COLUMNS,
 };
 use cdiscovery::{Config, Discovery};
+use cinformer::{handler::Handler, InformerEventSender, InformerService, MetaIoHandler, PubSubHandler, Session};
 use ckey::{Address, NetworkId, PlatformAddress};
 use ckeystore::accounts_dir::RootDiskDirectory;
 use ckeystore::KeyStore;
 use clap::ArgMatches;
 use clogger::{EmailAlarm, LoggerConfig};
 use cnetwork::{Filters, ManagingPeerdb, NetworkConfig, NetworkControl, NetworkService, RoutingTable, SocketAddr};
+use crossbeam::unbounded;
+use crossbeam_channel as crossbeam;
 use csync::snapshot::Service as SnapshotService;
 use csync::{BlockSyncExtension, BlockSyncSender, TransactionSyncExtension};
 use ctimer::TimerLoop;
@@ -52,6 +55,7 @@ fn network_start(
     cfg: &NetworkConfig,
     routing_table: Arc<RoutingTable>,
     peer_db: Box<dyn ManagingPeerdb>,
+    sender: InformerEventSender,
 ) -> Result<Arc<NetworkService>, String> {
     let addr = cfg.address.parse().map_err(|_| format!("Invalid NETWORK listen host given: {}", cfg.address))?;
     let sockaddress = SocketAddr::new(addr, cfg.port);
@@ -66,6 +70,7 @@ fn network_start(
         filters,
         routing_table,
         peer_db,
+        sender,
     )
     .map_err(|e| format!("Network service error: {:?}", e))?;
 
@@ -246,6 +251,19 @@ pub fn run_node(matches: &ArgMatches<'_>) -> Result<(), String> {
         panic_hook::set_with_email_alarm(email_alarm);
     }
 
+    // FIXME: unbound would cause memory leak.
+    // FIXME: The full queue should be handled.
+    // This will be fixed soon.
+    let (informer_connection_sender, informer_connection_receiver) = unbounded();
+    let informer_event_sender = {
+        if !config.informer.disable.unwrap() {
+            let (service, event_sender) = InformerService::new(informer_connection_receiver);
+            service.run_service();
+            event_sender
+        } else {
+            InformerEventSender::null_notifier()
+        }
+    };
     let pf = load_password_file(&config.operating.password_path)?;
     let base_path = config.operating.base_path.as_ref().unwrap().clone();
     let keys_path =
@@ -273,7 +291,14 @@ pub fn run_node(matches: &ArgMatches<'_>) -> Result<(), String> {
             let network_id = c.network_id();
             let routing_table = RoutingTable::new();
             let peer_db = PeerDb::new(c.get_kvdb());
-            let service = network_start(network_id, timer_loop, &network_config, Arc::clone(&routing_table), peer_db)?;
+            let service = network_start(
+                network_id,
+                timer_loop,
+                &network_config,
+                Arc::clone(&routing_table),
+                peer_db,
+                informer_event_sender,
+            )?;
 
             if config.network.discovery.unwrap() {
                 discovery_start(&service, &config.network, routing_table)?;
@@ -317,6 +342,17 @@ pub fn run_node(matches: &ArgMatches<'_>) -> Result<(), String> {
         self_nominate_start(c, matches, accountp, address);
     }
 
+    let informer_server = {
+        if !config.informer.disable.unwrap() {
+            let io: PubSubHandler<Arc<Session>> = PubSubHandler::new(MetaIoHandler::default());
+            let mut informer_handler = Handler::new(io);
+            informer_handler.event_subscription(informer_connection_sender);
+
+            Some(informer_handler.start_ws(config.informer_config())?)
+        } else {
+            None
+        }
+    };
     let (rpc_server, ipc_server, ws_server) = {
         let rpc_apis_deps = ApiDependencies {
             client: client.client(),
@@ -377,6 +413,9 @@ pub fn run_node(matches: &ArgMatches<'_>) -> Result<(), String> {
 
     wait_for_exit();
 
+    if let Some(server) = informer_server {
+        server.close_handle().close();
+    }
     if let Some(server) = rpc_server {
         server.close_handle().close();
         server.wait();
--- a/informer/src/handler/connection.rs
+++ b/informer/src/handler/connection.rs
@@ -1,0 +1,66 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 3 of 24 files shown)


informer/src/handler/connection.rs
```diff
@@ -0,0 +1,70 @@
...
+impl Connection{
+    pub fn new(status: bool, events: String , sink: Sink) -> Self {
+
+        Self{
+            status: status,
+            intrested_event: vec![events.clone()],
+            sink: sink.clone(),
+        }
+    }
+    pub fn event_subscription <T> (&mut self, mut io : PubSubHandler<T>) {
```
majecty: Will you add this function in the Connection struct? I feel awkward. The connection will be created whenever a user is connected. `io.add_subscription` will be called only once.

MSNTCS: I think `io.add_subscription` will be called whenever a user sends a subscription request. and will not be called once.

majecty: Since their documentation is not good, it is hard to know their intention.
With my common sense about webserver, what `io.add_subscription` is doing is registering a callback to the WebSocket server. In web server, we do not register callback multiple times.

In my assumption, the closure you passed to the `io.add_subscription`' will be called whenever user connects.

Could you check the `io.add_subscription`'s behavior?

informer/src/handler/connection.rs
```diff
@@ -0,0 +1,70 @@
...
+                    .reject(Error{
+                        code: ErrorCode::ParseError,
+                        message: "Invalid parameters. Subscription rejected.".into(),
+                        data: None,
+                    })
+                    .unwrap();
+                return;
+            }
+
+            thread::spawn(move || {
```
majecty: Why do we need a new thread here?

MSNTCS: This is for my experience of how we can manage a thread. do not mind it for now.

informer/src/handler/connection.rs
MSNTCS: `add_subscription` creates a new `subscribe` and `unsubscribe` RPC methods which their types are `SubscribeRpcMethod` and `UnsubscribeRpcMethod`. Then add these two so-called asynchronous methods to RPC request handler (`MetaIoHandler`). Therefore, I cannot conclude whether we need to create a subscription and unsubscription for each event by ourself (before anybody sends any subscription request) or we need to create a subscription/un per each request.

informer/src/handler/connection.rs
```diff
@@ -0,0 +1,65 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{Events, Params, Sink, Value};
+use futures::Future;
+use jsonrpc_core::futures;
+
+#[derive(Clone)]
```
majecty: When do you need to clone the Connection struct?

informer/src/informer_service/dummy_informer_service.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[derive(Clone, PartialEq)]
+pub enum Events {
+    None,
+    PeerAdded(String, String, usize),
+}
+#[derive(Clone)]
+pub struct InformerService {
+    connections: Vec<Connection>,
+    pub sender: NotifySender,
```
majecty: Why does the InformerService have a sender? I thought that the InformerService would have only receivers.

MSNTCS: Because I create a channel for sending connection in  `run node` and share with `handler` and `informer service`. There

majecty: IMO, since the sender is not used in the InformerService, it should not have it.
There are two options for how to remove the sender from the InfomreService struct.

One: making InformerService::new return the sender. For example
```rs
pub fn new(connection_receiver: Receiver<Connection>) -> (Self, NotifySender) {
```

Two: making the InformerService::new receive the sender. For example:
```rs
    pub fn new(connection_receiver: Receiver<Connection>, event_receiver: Receiver<Events>) -> Self {
```

MSNTCS: I got your point and am agree with you. I will use the first way.

informer/src/handler/connection.rs
MSNTCS: I need to clone the connections of InfromerService when the code is notifying all clients.

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+    pub fn event_subscription(&mut self, sender: Sender<Connection>) {
+        self.handler.add_subscription(
+            "Your event subscription begins",
+            ("register", move |params: Params, _, subscriber: Subscriber| {
+                let paramsclone = params.clone();
+                let unwraped_params: Result<Vec<String>, Error> = params.parse();
+                let allparams = unwraped_params.unwrap();
+                println!("####################");
```
MSNTCS: @majecty I forgot to change these, I will change it after your review is done.

informer/src/handler/connection.rs
```diff
@@ -0,0 +1,67 @@
...
+                let event_params: Vec<Value> = vec![message.into(), address.into(), token.into()];
+                match self.sink.notify(Params::Array(event_params)).wait() {
+                    Ok(_) => {}
+                    Err(_) => {
+                        cinfo!(INFORMER, "Subscription has ended, finishing.");
+                    }
+                }
+            }
+            _ => cinfo!(INFORMER, "invalid Event: the event is not supported"),
+        }
```
majecty: You can make the code simple using serde.

```
let json_object = serde_json::to_value(event).unwrap().as_object();
let params = Params::Map(json_object);
match self.sink.notify(Params::Array(event_params)).wait() {
    Ok(_) => {}
    Err(_) => {
        cinfo!(INFORMER, "Subscription has ended, finishing.");
    }   
}
```

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::io;
+use std::sync::Arc;
+
+pub struct InformerConfig {
+    pub interface: String,
+    pub port: u16,
+    pub max_connections: usize,
+}
+
+#[derive(Default)]
```
majecty: Why do you need `Default`?

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+    pub fn start_ws(self, config: InformerConfig) -> Result<WsServer, String> {
+        let url = format!("{}:{}", config.interface, config.port);
+        let addr = url.parse().map_err(|_| format!("Invalid WebSockets listen host/port given: {}", url))?;
+        let server = self.handler;
+        let start_result = start_ws(&addr, server, config.max_connections);
+        match start_result {
+            Err(WsError::Io(ref err)) if err.kind() == io::ErrorKind::AddrInUse => {
+                Err(format!("WebSockets address {} is already in use, make sure that another instance of a Codechain node is not running or change the address using the --ws-port options.", addr))
```
majecty: Is informer using `--ws-port` option?

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+            },
+            Ok(server) => {
+                cinfo!(INFORMER, "WebSockets Listening on {}", addr);
+                Ok(server)
+            },
+        }
+    }
+    pub fn event_subscription(&mut self, sender: Sender<Connection>) {
+        self.handler.add_subscription(
+            "Your event subscription begins",
```
majecty: What message should we use here?

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+            ("register", move |params: Params, _, subscriber: Subscriber| {
+                let paramsclone = params.clone();
+                let unwraped_params: Result<Vec<String>, Error> = params.parse();
+                let allparams = unwraped_params.unwrap();
```
majecty: You should not `unwrap` here. When a user gives wrong parameter, Foundry should not panic.

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+                cinfo!(INFORMER, "WebSockets Listening on {}", addr);
+                Ok(server)
+            },
+        }
+    }
+    pub fn event_subscription(&mut self, sender: Sender<Connection>) {
+        self.handler.add_subscription(
+            "Your event subscription begins",
+            ("register", move |params: Params, _, subscriber: Subscriber| {
+                let paramsclone = params.clone();
```
majecty: Why did you clone params here?

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+                println!("####################");
+                println!("{}", allparams[0]);
+                if paramsclone == Params::None {
+                    subscriber
+                        .reject(Error {
+                            code: ErrorCode::ParseError,
+                            message: "Invalid parameters. Subscription rejected.".into(),
+                            data: None,
+                        })
+                        .unwrap();
```
majecty: Do you know when the `unwrap` fails? Please use `expect` rather than `unwrap` and use a message that contains why you convinced that this will not happen.

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,103 @@
...
+                let mut rng = rand::thread_rng();
+                let sub_id = rng.gen();
+                println!("{}", sub_id);
+                let sink = subscriber.assign_id(SubscriptionId::Number(sub_id)).unwrap();
+                match sink.notify(Params::Array(vec![Value::String("You are now a new subscriber".into())])).wait() {
+                    Ok(_) => {}
+                    Err(_) => {
+                        cinfo!(INFORMER, "Subscription has ended, finishing.");
+                    }
+                }
```
majecty: Is this needed?

foundry/config/mod.rs
```diff
@@ -125,6 +128,16 @@ impl Config {
         }
     }
 
+    pub fn informer_config(&self) -> InformerConfig {
+        debug_assert!(!self.informer.disable.unwrap());
```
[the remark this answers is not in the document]
MSNTCS: I followed the same pattern as other functions.

```
pub fn rpc_ipc_config(&self) -> RpcIpcConfig {
        debug_assert!(!self.ipc.disable.unwrap());

        RpcIpcConfig {
            socket_addr: self.ipc.path.clone().unwrap(),
        }
    }
```

informer/src/handler/rpc_ws_handler.rs
MSNTCS: here is what a user received.
```
Received:  {"jsonrpc":"2.0","method":"Your event subscription begins","params":["You are now a new subscriber"]}
```

informer/src/handler/rpc_ws_handler.rs
MSNTCS: The purpose of this message is to let the client know that the subscription successfully occurs.
here is what the client receives after a successful subscription.
```
Received:  {"jsonrpc":"2.0","result":14308347713743523176,"id":1}
Received:  {"jsonrpc":"2.0","method":"Your event subscription begins","params":["You are now a new subscriber"]}
```

foundry/run_node.rs
```diff
@@ -317,6 +348,18 @@ pub fn run_node(matches: &ArgMatches<'_>) -> Result<(), String> {
         self_nominate_start(c, matches, accountp, address);
     }
 
+    let mut _informer_handler = Handler::default();
```
majecty: Why did you add prefix "_" in the variable name?

foundry/run_node.rs
```diff
@@ -317,6 +348,18 @@ pub fn run_node(matches: &ArgMatches<'_>) -> Result<(), String> {
...
+    let _informer_server = {
```
majecty: ditto.

majecty:
Please use `expect` instead of `unwrap`.
We can use `unwrap` only when the situation is obvious.

informer/src/handler/rpc_ws_handler.rs
majecty: We don't need it. Do you know how to prohibit the library from sending the message?

informer/src/informer_service/informer_notify.rs
```diff
@@ -0,0 +1,53 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+use crate::Events;
+use crossbeam::{bounded, unbounded, Receiver, Sender};
+use crossbeam_channel as crossbeam;
+
+pub fn create() -> (NotifySender, Receiver<Events>) {
+    let (tx, rx) = unbounded();
```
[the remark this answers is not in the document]
MSNTCS: I add comments

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,94 @@
...
+                let unwrapped_params: Result<Vec<String>, Error> = params.parse();
+                let all_params =
+                    unwrapped_params.expect("The parameter that the client sent for subscription is not valid");
+                let mut rng = rand::thread_rng();
+                let sub_id = rng.gen();
+                let sink = subscriber
+                    .assign_id(SubscriptionId::Number(sub_id))
+                    .expect("Subscription request is already terminated");
+                let mut connection = Connection::new(true, sink, sub_id);
```
majecty: It's hard to know what it is when you just pass the `true` value.
I have two suggestions.
* Use an enum for it like "ConnectionState::Connected" and "ConnectionState::Disconnected".
* Do not receive the state from the `new` function. You will do not want to create a connection with the state value `false`.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -0,0 +1,102 @@
...
+            cinfo!(INFORMER, "Informer Service has started");
+            let mut select = crossbeam::Select::new();
+            let event_index = select.recv(&event_rcv);
+            let connection_index = select.recv(&connection_rcv);
+            loop {
+                let mut s = select.clone();
+                match s.ready() {
+                    index if index == event_index => {
+                        if let Ok(event) = event_rcv.try_recv() {
+                            cinfo!(INFORMER, "Event is send to all clients");
```
majecty:
```suggestion
                            cinfo!(INFORMER, "Event is sent to all clients");
```

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -0,0 +1,102 @@
...
+                            let popup_event = event;
```
majecty: Why did you create a new variable?

informer/src/handler/connection.rs
```diff
@@ -0,0 +1,60 @@
...
+                cinfo!(INFORMER, "The event is successfully added to user's interested events");
+                self.intrested_events.push(event);
+            }
+            _ => {
+                cinfo!(INFORMER, "invalid Event: the event is not supported");
+            }
+        }
+    }
+
+    pub fn notify_client(&mut self, event: &Events) {
```
majecty: You can remove `mut` from the `notify_client` function like this:
```
diff --git a/informer/src/handler/connection.rs b/informer/src/handler/connection.rs
index d2b678e4..8d6c085e 100644
--- a/informer/src/handler/connection.rs
+++ b/informer/src/handler/connection.rs
@@ -47,7 +47,7 @@ impl Connection {
         }
     }
 
-    pub fn notify_client(&mut self, event: &Events) {
+    pub fn notify_client(&self, event: &Events) {
         let json_object = serde_json::to_value(event).unwrap().as_object_mut().cloned();
         let params = Params::Map(json_object.expect("json format is not valid"));
         match self.sink.notify(params).wait() {
diff --git a/informer/src/informer_service/informer_service_handler.rs b/informer/src/informer_service/informer_service_handler.rs
index e40e3aa2..32f8a3c2 100644
--- a/informer/src/informer_service/informer_service_handler.rs
+++ b/informer/src/informer_service/informer_service_handler.rs
@@ -90,8 +90,8 @@ impl InformerService {
         }
     }
 
-    pub fn notify_client(&mut self, popup_event: Events) {
-        for connection in &mut self.connections {
+    pub fn notify_client(&self, popup_event: Events) {
+        for connection in &self.connections {
             for interested_event in connection.intrested_events.clone() {
                 if InformerService::compare_event_types(&interested_event, &popup_event) {
                     connection.notify_client(&popup_event);

```

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -0,0 +1,102 @@
...
+        let event_rcv = self.event_receiver.clone();
+        let connection_rcv = self.connection_receiver.clone();
+        spawn(move || {
+            cinfo!(INFORMER, "Informer Service has started");
+            let mut select = crossbeam::Select::new();
+            let event_index = select.recv(&event_rcv);
+            let connection_index = select.recv(&connection_rcv);
+            loop {
+                let mut s = select.clone();
+                match s.ready() {
```
majecty:
```suggestion
                match select.ready() {
```

network/src/p2p/handler.rs
```diff
@@ -120,8 +121,11 @@ pub struct Handler {
     max_peers: usize,
     peer_db: Box<dyn (ManagingPeerdb)>,
     rng: Mutex<OsRng>,
+
+    event_sender: NotifySender,
```
majecty: The names`NotifySender` and `event_sender` are too general.
It is hrad to know what they are using the name.
How about `informer_sender: InformerSender` or `informer_event_sender: InformerEventSender`?

informer/src/handler/rpc_ws_handler.rs
```diff
@@ -0,0 +1,94 @@
...
+                            code: ErrorCode::ParseError,
+                            message: "Invalid parameters. Subscription rejected.".into(),
+                            data: None,
+                        })
+                        .expect("Invalid parameters are used. Subscription rejected");
+                    return
+                }
+                let unwrapped_params: Result<Vec<String>, Error> = params.parse();
+                let all_params =
+                    unwrapped_params.expect("The parameter that the client sent for subscription is not valid");
```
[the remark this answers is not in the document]
MSNTCS: This error would not be for the client. this error is for foundry.

PR #218, opened 2020-03-02
Implement ICS 04

junha1:
There are still missing required functions, since we don't need it in PoC

PR #221, opened 2020-03-02
Implement Datagram & transaction handler for ICS04

junha1:
Depends on https://github.com/CodeChain-io/foundry/pull/218

### DIFF
--- a/core/src/ibc/transaction_handler/mod.rs
+++ b/core/src/ibc/transaction_handler/mod.rs
@@ -132,5 +132,6 @@ pub fn execute(
                 .handle_open_confirm(identifier, proof_ack, proof_height)
                 .map_err(|err| RuntimeError::IBC(format!("ConnOpenConfirm: {}", err)).into())
         }
+        _ => unimplemented!(),
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/ibc/transaction_handler/mod.rs
```diff
@@ -132,6 +133,113 @@ pub fn execute(
...
+                            ibc::channel_04::types::ChannelOrder::UNORDERED
+                        }
+                    },
+                    raw.connection,
+                    raw.channel_identifier,
+                    raw.counterparty_channel_identifier,
+                    raw.version,
+                )
+                .map_err(|err| RuntimeError::IBC(format!("ChanOpenInit: {}", err)).into())
+                .map(|_| ())
```
majecty: Why did you add ` .map(|_| ())` here?
How about this?

```
            channel_manager
                .chan_open_init(
                    {
                        if raw.order == 1 {
                            ibc::channel_04::types::ChannelOrder::ORDERED
                        } else {
                            ibc::channel_04::types::ChannelOrder::UNORDERED
                        }
                    },
                    raw.connection,
                    raw.channel_identifier,
                    raw.counterparty_channel_identifier,
                    raw.version,
                )
                .map_err(|err| RuntimeError::IBC(format!("ChanOpenInit: {}", err)).into())?;
            Ok(())
```

junha1: That explicitly discards any returned result. I have no opinion about that, so I'll just follow your suggestion.

majecty:
Good. Let's merge this PR after #218.

PR #222, opened 2020-03-02
Create IBC light clients in IBC scenario

majecty:


ibc.ts/src/scenario/index.ts
```diff
@@ -1,7 +1,87 @@
...
+    debug("Create light client");
+    const counterpartyBlockNumber = await counterpartyChain.latestHeight();
+    const blockNumber = await chain.latestHeight();
+    debug(`height is ${counterpartyBlockNumber}`);
+    const counterpartyRawHeader = await counterpartyChain.queryChainHeader(
+        counterpartyBlockNumber
+    );
+    debug(`rawHeader is ${counterpartyRawHeader}`);
+
+    assert(counterpartyRawHeader, "headet should not empty");
```
junha1: "header should not be empty"

majecty: @junha1 I fixed it.

junha1: @majecty  not 'be'?

majecty: @junha1 I fixed it.

PR #226, opened 2020-03-03
Implement ICS04 RPCs

junha1:


core/src/ibc/channel_04/manager.rs
```diff
@@ -15,7 +15,9 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
majecty: What do you mean by the commit message "Refactor ICS data that are stored in hash"?
Is "in hash" commonly used term?

junha1: `in hash` was meant to be `as a hash`. @ScarletBlue answered that `as a hash` would be more natural. I'll update it

majecty:
@junha1 Could you fix the conflict?

junha1:
@majecty done

PR #229, opened 2020-03-03
[WIP] Implement.execute.block.call

Byeongjee:
I tried to simplify the work as far as I could.

TODO:
- [x] Use `execute_block` interface in importer
- [ ] Implement and use `commit` and `revert` interface
- [ ] Decouple double vote reporting logic from host
    - [ ] Create double vote type in coordinator crate (cannot import from the core because of cyclic dependency)
- [ ] Change transaction encoding
    - [ ] Do less decoding in `VerificationQueue`
- [ ] Inject coordinator reference to engine
- [ ] Implement `execute_block`
    - transaction verification should be done here
Remaining issues:
- Handling `evidence` when calling `execute_block`
- where should we call `mempool.remove_old`?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,6 +327,7 @@ dependencies = [
  "codechain-timer",
  "codechain-types",
  "codechain-vm",
+ "coordinator",
  "crossbeam-channel 0.3.8",
  "hyper 0.10.0-a.0",
  "kvdb",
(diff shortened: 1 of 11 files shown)


Byeongjee:
There was mistake. I'm rewriting the requirement

Byeongjee:
Remaining work should be done after we decide our evidence handling logic, encoding scheme, or coordinator injection logic. I'll work on `check_transaction` for now.

Byeongjee:
merged into https://github.com/CodeChain-io/foundry/pull/260

PR #231, opened 2020-03-04
Fix relayer to update each chain's light client successfully

majecty:


core/src/ibc/transaction_handler/mod.rs
```diff
@@ -59,7 +59,7 @@ pub fn execute(
             header,
```
junha1: Make the UpdateClient function use the correnct method name in log
->
Make the UpdateClient function use the correct method name in log

core/src/consensus/light_client/verification.rs
```diff
@@ -59,13 +59,21 @@ pub fn verify_quorum(vset: &CompactValidatorSet, seal: &Seal) -> bool {
...
     if client_state.number + 1 != proposal.number {
         return false
     }
     if client_state.next_validator_set_hash != proposal.validator_set.hash() {
         return false
     }
+
+    // FIXME: We should remove this if statement when the static validator is removed.
+    if BYPASS_VERIFICATION_IN_STATIC_VALIDATOR {
+        return true
```
junha1: Was this ok with clippy?

ibc.ts/chainA/chainA.schem.json
```diff
@@ -6,7 +6,7 @@
         "validators": [
           "0xc7db8f558a87a45d3b4d2e7422abb48a27e6eeee00a0f884ca244d1d783d6ead2ca445e9ac8f08b311d78b4b4b69d48a9daf0c98d2a19e87e482f8060138ba88"
         ],
-        "timeoutPropose": 3000,
+        "timeoutPropose": 10000,
```
junha1: Relayer can't handle one block per 3 sec?

core/src/miner/miner.rs
```diff
@@ -328,6 +328,9 @@ impl Miner {
         };
 
         assert!(self.engine.seals_internally(), "If a signer is not prepared, prepare_block should not be called");
+        // FIXME this sleeping is for the experimental branch. This should not be included in the master branch.
```
junha1: In master, instead of this, we will fix the tendermint core logic?

ibc.ts/src/common/chain.ts
```diff
@@ -54,17 +54,34 @@ export class Chain {
     }
```
junha1: If we send transactions without waiting for inclusion, Foundy can
create a block that has more than one transaction.
->
If we send transactions without waiting for inclusion, Foundry can
create a block that has more than one transaction.

junha1:
Please check cancelled CI actions

ibc.ts/chainA/chainA.schem.json
majecty: Relayer can handle it. I wrote the reason in the commit message. Wasn't it enough?
If you thought that some information was missing in the commit message, please tell me the point.

Below is the commit message I copied from the commit.
> Relayer should run harder to update light clients if the chain creates
block too fast.

core/src/miner/miner.rs
majecty: We don't need it.
Since network latency, a Foundry used in a network does not need this code.

ibc.ts/chainA/chainA.schem.json
junha1: "Should run harder" sounds like that it can't handle in some cases. If it *can* handle but just harder, what's the problem?

majecty: Thanks for your comment. Relayer will use more time to update the light client. I'll fix the commit's description.

junha1: You mean Relayer becomes the bottleneck?

majecty: I don't understand what you mean bottleneck. Let's have a call.

core/src/consensus/light_client/verification.rs
majecty: Yes, it passed.

junha1:
@majecty Why do some of CI keep fail?

PR #232, opened 2020-03-04
Build a connection

majecty:
This PR depends on https://github.com/CodeChain-io/foundry/pull/231

- [x] Send ConnOpenInit
- [x] Send ConnOpenTry
- [x] Send ConnOpenAck
- [x] Send ConnOpenConfirm


ibc.ts/src/relayer/index.ts
```diff
@@ -120,7 +142,8 @@ async function updateLightClient({
         );
     }
     let currentBlockNumber = clientState!.data!.number;
-    while (currentBlockNumber < counterpartyChainHeight) {
+    // We can't get the best block's IBC header
```
junha1: I think `FIXME` is needed here, because there is `FIXME` in `compose_header()` for the same problem.

ibc.ts/src/relayer/index.ts
```diff
@@ -100,6 +100,10 @@ async function pendingDatagrams({
         })
     );
 
+    // We can't update light client upto the best block.
+    height = height - 1;
```
junha1: ditto

majecty:
@junha1 I fixed them.

PR #234, opened 2020-03-05
[WIP] Implement check_transaction call

Byeongjee:
We decide to remove `FutureQueue` from mempool, so TODO has been changed.
TODO:
- [x] Remove `FutureQueue` from mempool
- [x] Implement `mem_pool.add` using `check_transaction`
~~Implement future queue update method~~
- [x] Replace `TransactionOrder` type with `Priority`
- [x] Change callers of the mempool methods
- ~~[ ] Design and use `create_block` interface~~
 - ~~This should filter away old Txes from mempool and return a list of ordered Txes to be included in a block (ordering should be valid)~~
- ~~[ ] Design and use an interface to support removing invalid transactions (`remove_old`)~~
- ~~This should be similar to the `create_block` method. It should filter away old Txes from mempool, after accepting each block.~~
- [x] Design and use `fetch_transactions_for_block` interface
- This will return a list of ordered transactions that can be included in a block
- This will not filter away old Txes
- Miner does not have to include all returned transactions into a block. It will be the miner's responsibility to enforce the gas/size limit of blocks.
- [x] Design and use 'remove_old_transactions` interface
- This should be called after accepting each block.
- [ ] Change transaction encoding scheme

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,6 +327,7 @@ dependencies = [
  "codechain-timer",
  "codechain-types",
  "codechain-vm",
+ "coordinator",
  "crossbeam-channel 0.3.8",
  "hyper 0.10.0-a.0",
  "kvdb",
--- a/coordinator/src/context.rs
+++ b/coordinator/src/context.rs
@@ -1,2 +1,7 @@
 /// A `Context` provides the interface against the system services such as
 pub trait Context {}
+
+#[derive(Default)]
+pub struct DummyContext {}
+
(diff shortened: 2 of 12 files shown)


Byeongjee:
We decided to remove `FutureQueue` from `mempool`. mempool will reject future transactions and it will be the application's responsibility to resend the rejected transactions.

Byeongjee:
I merged this work into https://github.com/CodeChain-io/foundry/pull/260

PR #235, opened 2020-03-06
Fix light client

junha1:
Previous algorithm has a critical problem.

### DIFF
--- a/core/src/consensus/light_client/mod.rs
+++ b/core/src/consensus/light_client/mod.rs
@@ -44,10 +44,11 @@ impl ClientState {
 /// All data for updating a light client up to Nth block.
 #[derive(RlpEncodable, RlpDecodable, PartialEq, Debug)]
 pub struct UpdateHeader {
-    /// 'N'
-    pub number: BlockNumber,
-    /// Hash of Nth block.
-    pub hash: H256,
+    /// header.BlockNumber will be 'N'
+    /// TODO: we can carry only required fields if we elaborate our current block hashing scheme
+    pub header_raw: Header,
+
+    /// These two are not part of the actual block header.
     /// Seal to Nth block which will be stored in (N+1)th block.
     pub seal: Seal,
     /// Validator set of Nth block which will be stored in (N-1)th state.
@@ -57,8 +58,8 @@ pub struct UpdateHeader {
 impl UpdateHeader {
     pub fn make_client_state(&self) -> ClientState {
         ClientState {
-            number: self.number,
-            next_validator_set_hash: self.validator_set.hash(),
+            number: self.header_raw.number(),
+            next_validator_set_hash: *self.header_raw.next_validator_set_hash(),
         }
     }
 }
--- a/core/src/ibc/client_02/client.rs
+++ b/core/src/ibc/client_02/client.rs
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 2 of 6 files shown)


rpc/src/v1/impls/ibc.rs
```diff
@@ -102,28 +102,33 @@ where
...
-        let state = self.client.state_at(block_id).unwrap();
-
+        if self.client.state_at(block_id_next).is_none() {
+            return Ok(None)
+        }
+        let state_prev = self.client.state_at(block_id_prev).unwrap();
         let header_core = self.client.block_header(&block_id).unwrap();
-        let vset_raw = ccore::stake::NextValidators::load_from_state(&state).map_err(errors::core)?;
+        let header_core_next = self.client.block_header(&block_id_next).unwrap();
+        let vset_raw = ccore::stake::NextValidators::load_from_state(&state_prev).map_err(errors::core)?;
```
majecty: If we use `CurrentValidators`, we don't need to read the previous block.

rpc/src/v1/impls/ibc.rs
```diff
@@ -102,28 +102,33 @@ where
     }
 
     fn compose_header(&self, block_number: u64) -> Result<Option<Bytes>> {
+        // FIXME: Currently compose_header() works only for at most bestblock-1
```
majecty: Could you give more context?
Like "block n's seal is stored in the next block's header".

rpc/src/v1/impls/ibc.rs
junha1: Ok, then I'll add `create_compact_validator_set()` on `CurrentValidators` and use that.

rpc/src/v1/impls/ibc.rs
junha1: Ok

junha1:
@majecty Updated.

core/src/consensus/light_client/mod.rs
```diff
@@ -44,10 +44,11 @@ impl ClientState {
 /// All data for updating a light client up to Nth block.
```
majecty: For the " Fix invalid light client core algorithm " commit, please write these things in your commit message:
* What the problem was: number was not verified.
* How you fixed the problem: include a raw header in the "UpdateHeader".

core/src/ibc/client_02/client.rs
```diff
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
```
majecty: For the commit message: " Adapt ICS02 client to the new core light client algorithm", I don't understand the meaning of the message. Could you explain it to me?

rpc/src/v1/impls/ibc.rs
```diff
@@ -102,21 +102,29 @@ where
     }
```
majecty: For the commit message "Fix to provide correct seal and validator set", please write what was the problem and how you fixed it.

core/src/ibc/client_02/client.rs
junha1: @majecty 
Means ICS 02 has to be changed along the change of core light client algorithm.

majecty: Did you mean that we should change our ICS02 light client implementation along to the change of the light client you changed at the previous commit?
If it is right, please merge it with the light client commit or give a brief description about what was changed in the light client code.

junha1: I think it's better to just squash

junha1:
@majecty  Updated

PR #237, opened 2020-03-06
Fit test framework types into foundry-keystore

HoOngEe:
It can be merged only after the #154


HoOngEe:
Only the `d18c3bc` is added onto the #154

Issue #7, opened 2019-12-19
Remove Assets


majecty:
Each module will define its assets. We don't need to provide a general asset mechanism in Foundry.

- [x] Remove asset-related RPCs
- [x] Remove asset-related transactions
- [x] Remove the types related to asset from `types`
    - TransferAsset
        - struct AssetTransferInput
        - struct AssetOutPoint
        - struct AssetTransferOutput
    - struct AssetMintOutput    
    - struct AssetWrapCCCOutput
- [x] Remove assets from MemPoolMinFees in `miner` and `rpc types`
- [x] Remove assets from the e2e tests
- [x] Modify the e2e tests

somniumism:
PR #246 resolved this issue.

HoOngEe:
Let's remove the `Asset` also from documents, test macro, json scheme texts

PR #246, opened 2020-03-10
Remove assets from the foundry

somniumism:
- [x] Remove the types related to the asset from `types`
    - TransferAsset
        - struct AssetTransferInput
        - struct AssetOutPoint
        - struct AssetTransferOutput
    - struct AssetMintOutput    
    - struct AssetWrapCCCOutput
- [x] Remove assets from MemPoolMinFees in `miner` and `rpc types`
- [x] Remove assets from the e2e tests
- [x] Modify the e2e tests

resolves issue #7


somniumism:
@HoOngEe I added you as a reviewer. There may be some parts that I forgot to remove. Therefore, please clone this PR and review carefully : )

I didn't remove `AssetTransaction` in `test/src/sdk/core/Transaction.ts` and the logic related to it, because It seems necessary for the transactions. Please check this part, too!

kseo:
@somniumism this PR doesn't need to be complete. You can create following PRs if you missed something in this PR.

vm/src/executor.rs
```diff
@@ -77,10 +73,12 @@ impl From<HashingError> for RuntimeError {
 struct Item(Vec<u8>);
 
 impl Item {
+    #[allow(dead_code)]
```
HoOngEe: I think you'd better to write higher level warning suppression.
You can write `#[allow(dead_code)]` right above the mod executor in `vm/src/lib.rs` or `#![allow(dead_code)]` for crate level suppression. https://doc.rust-lang.org/rust-by-example/attribute.html

HoOngEe:
@somniumism We can remove the `AssetTransaction` also. We do not manage asset layer any longer. Let's remove it. Other parts are looking good

vm/src/executor.rs
somniumism: Thank you for your good idea! I modified it : ))

PR #247, opened 2020-03-10
Build a channel

majecty:
I wrote code that builds a channel between two chains.
~I found that there is an error while verifying the connection hops. I'll fix it soon after reading the channel spec again.~
I fixed all the errors.

### DIFF
--- a/core/src/ibc/channel_04/manager.rs
+++ b/core/src/ibc/channel_04/manager.rs
@@ -54,12 +54,17 @@ impl<'a> Manager<'a> {
     }
 
     // Utility functions
-    fn check_connection_opened(&self, id: IdentifierSlice) -> Result<Identifier, String> {
+    fn query_connection(&self, connection_id: IdentifierSlice) -> Result<ConnectionEnd, String> {
         let kv_store = self.ctx.get_kv_store();
-        let connection_end: ConnectionEnd =
-            rlp::decode(&kv_store.get(&connection_path(&id)).ok_or_else(|| "Connection doesn't exist".to_owned())?)
-                .expect("Illformed ConnectionEnd stored in the DB");
+        let connection_end: ConnectionEnd = rlp::decode(
+            &kv_store.get(&connection_path(&connection_id)).ok_or_else(|| "Connection doesn't exist".to_owned())?,
+        )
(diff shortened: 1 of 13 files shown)


core/src/ibc/channel_04/manager.rs
```diff
@@ -203,14 +208,15 @@ impl<'a> Manager<'a> {
...
+        let connection_end = self.query_connection(&connection)?;
 
         let expected = ChannelEnd {
             state: ChannelState::INIT,
             ordering: order,
             counterparty_port_identifier: DEFAULT_PORT.to_string(),
             counterparty_channel_identifier: channel_identifier.clone(),
             // Note: the array should be reversed in the future where `connection` becomes an array.
-            connection_hops: vec![connection.clone()],
+            connection_hops: vec![connection_end.counterparty_connection_identifier],
```
junha1: Note that ICS javascript example uses just `connection` (given as an argument) here. It's ok not to change if you think this is not important.

core/src/ibc/channel_04/manager.rs
```diff
@@ -54,12 +54,17 @@ impl<'a> Manager<'a> {
     }
```
junha1: What is "Acked-by: Park Juhyung <jh@codechain.io>" in your commit message? ('Fix connection hop to use the same chain's identifier')

majecty: I added it by mistake. I'll remove it.

core/src/ibc/channel_04/manager.rs
majecty: Since the word "connection" was used in the channel/manager.rs code for the connection identifier, I used the term "connection_end".

We need a consistent naming rule. What do you prefer? Shall we use `connection` for the identifier? Shall we use `connection` for the ConnectionEnd?

junha1: Ah, I was misunderstanding something. The name `connection` is already used in the function parameter, which represents `connectionHops` in javascript. And I misunderstood that we use `previous` here. 
About your point, yes you're right but I think we can just stick to our convention (`connection` for id and `connection_end` for struct).

PR #248, opened 2020-03-11
Implement basic IPC, synchronization, and execution for sandbox

junha1:


basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,75 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+
+use crate::ipc::semaphore::{Mode, Semaphore};
+use crate::ipc::Ipc;
+use std::process::Command;
+
+const SOCKET_SRC: &str = "./Foundry_Socket_Boxer";
+const SOCKET_DST: &str = "./Foundry_Socket_Boxee";
```
dynaxis: Even though the paths are all relative to the current working directory, I think they may be problemsome when we create multiple sandboxes. If so, I think it's a safe bet to generate the name from `Uuid::new_v4().to_simple()` by leveraging `uuid` crate.

junha1:
@majecty I requested you review since there are some changes on Foundry's project configurations.

testmodule/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+# How to add a new one (Python)
+Not supported yet
+
+# How to add a new one (TypeScript)
+Not supported yet
+
+# How to add a new one (WASM Binary)
+Not supported yet
+
```
majecty: README should start with the first level header.
Please only one first level header in the README.md file.

.github/workflows/cargo-test.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           override: true
       - run: cargo fetch --verbose
       - run: cargo build
+      - run: cargo build --package codechain-testmodule
```
majecty: Why do we need this line?
`cargo build` does not build codechain-testmodule?

testmodule/src/simple.rs
```diff
@@ -0,0 +1,12 @@
...
+use cbasesandbox::ipc::Ipc;
+
+#[cfg(all(unix, target_arch = "x86_64"))]
+fn main() -> Result<(), String> {
+    let mut ctx = executee::start::<cbasesandbox::execution::IpcUnixDomainSocket>();
+    let r = ctx.ipc.recv();
+    println!("CLIENT recv: {:?}", String::from_utf8(r));
+    ctx.ipc.send(b"I'm here!\0");
+    Ok(())
+}
```
majecty: Reviewing what this code is doing seems not my role. Please request a review to others. Maybe seulgi?

.github/workflows/cargo-test.yml
junha1: No it doesn't since `codechain-testmodule` is not used anywhere.

testmodule/README.md
junha1: Ok thanks

testmodule/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "codechain-testmodule"
+version = "0.1.0"
+authors = ["CodeChain Team <hi@codechain.io>"]
+edition = "2018"
+
+[dependencies]
+codechain-basesandbox = { path = "../basesandbox" }
+nix = "0.17.0"
+
+[[bin]]
```
junha1: @majecty This is ok?

.github/workflows/cargo-test.yml
majecty: If the `cargo build` does not build codechain-testmodule since it is not used in the Foundry binary, how about trying `cargo build --workspace`?

>--workspace
    Build all members in the workspace.

You can find a help message of `cargo build` here: https://doc.rust-lang.org/cargo/commands/cargo-build.html

junha1: Yes I've usually used this in local.
Then
```
- run: cargo build
- run: cargo build --package codechain-testmodule
```
to
```
- run: cargo build --workspace
```
?

majecty: Yes, it looks better to me.

testmodule/Cargo.toml
majecty: Does the test module need executable? Since I don't know the background, discussing it with another team member who knows the mold well is better.

junha1: @sgkim126 Any comments?

sgkim126: I think we need the binary. It's to test the module that creates a new process.
However, I'm not sure adding a binary to this repository is the best option yet.

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        Executor {
+            child: command.spawn().unwrap(),
+        }
+    }
+}
+
+impl Drop for Executor {
+    fn drop(&mut self) {
+        self.child.wait().unwrap();
```
sgkim126: Is it okay not to kill the child process and just wait?

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+/// declaration order of fields is important because of Drop dependencies
+pub struct Context<T: Ipc> {
+    pub ipc: T,
+    pub semaphore: Semaphore,
+    #[allow(dead_code)] // Executor implements Drop
+    child: Executor,
```
sgkim126: Renaming it to `_child` is the better way to show that the child is not explicitly used.

basesandbox/src/execution/executee.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::ipc::Ipc;
+
+// Interface for the sandboxee written in Rust
+pub struct Context<T: Ipc> {
+    pub ipc: T,
+    pub semaphore: Semaphore,
+}
+
+pub fn start<T: Ipc>() -> Context<T> {
+    println!("{:?}", std::time::SystemTime::now());
```
sgkim126: Printing the current time without any comments will not help users.

junha1: Ah it was for the debug. I'll remove it.

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+
+pub mod domain_socket;
+pub mod semaphore;
+
+pub trait Ipc: Drop {
+    /// This distinction may be required for some asymmetry IPC schemes
```
sgkim126: Do we have a need for an asymmetry IPC?
How about distinguishing a server and a client when we really need an asymmetric IPC?

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,67 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+
+use super::Ipc;
+use std::os::unix::net::UnixDatagram;
+use std::path::Path;
+
+pub struct DomainSocket {
+    pub address_src: String,
```
sgkim126: Why do you make it public?

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,75 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+
+use crate::ipc::semaphore::{Mode, Semaphore};
+use crate::ipc::Ipc;
+use std::process::Command;
+
+const SOCKET_SRC: &str = "./Foundry_Socket_Boxer";
```
sgkim126: Creating a file in the current directory doesn't seem good. How about using tmp directory?

basesandbox/src/execution/executor.rs
junha1: I think it depends on Foundry's overall module error handling policy. If the parent needs to kill the child explicitly, then it would indicate that something abnormal happened to child, since we expect it to terminate itself by some protocol.

basesandbox/src/execution/executee.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    let socket_src_name = &args[1];
+    let socket_dst_name = &args[2];
+    let semaphore_name = &args[3];
+
+    let mut semaphore = Semaphore::new(semaphore_name.to_string(), Mode::OPEN);
+
+    let mut ipc = T::create_server(socket_src_name.to_string(), socket_dst_name.to_string());
+    semaphore.signal();
+    assert!(ipc.recv() == b"#INIT");
```
sgkim126: Why not use `assert_eq`?

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,75 @@
...
+const SOCKET_DST: &str = "./Foundry_Socket_Boxee";
+const SEMAPHORE: &str = "/Foundry_Semaphore";
+
+struct Executor {
+    child: std::process::Child,
+}
+
+impl Executor {
+    fn new(path: &str, args: &[&str]) -> Self {
+        let mut command = Command::new(path);
```
sgkim126: Why not use [`args`](https://doc.rust-lang.org/std/process/struct.Command.html#method.args)?

basesandbox/src/ipc/mod.rs
junha1: Ah there was a mistake. `executee::start()` should call `T::create_client`. 
For example socket stream (not datagram) will differently implement those two.

basesandbox/src/ipc/semaphore.rs
```diff
@@ -0,0 +1,80 @@
...
+pub struct Semaphore {
+    raw: *mut libc::sem_t,
+    mode: Mode,
+    name: String,
+}
+
+impl Semaphore {
+    pub fn new(name: String, mode: Mode) -> Self {
+        unsafe {
+            if let Mode::CREATE = mode {
```
sgkim126:
```rust
if Mode::CREATE == mode {
```

basesandbox/src/ipc/semaphore.rs
```diff
@@ -0,0 +1,80 @@
...
+use nix::errno::Errno;
+use nix::libc;
+
+pub enum Mode {
+    CREATE,
+    OPEN,
+}
+
+/// Inter-process semaphore using POSIX
+pub struct Semaphore {
```
sgkim126: Do we need to implement a semaphore? Isn't there any library for it?

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,67 @@
...
+        socket: UnixDatagram::bind(address_src).unwrap(),
+        buffer: vec![0; 1024],
+    }
+}
+
+/// No distinction between server / client for DomainSocket in Drop
+impl Drop for DomainSocket {
+    fn drop(&mut self) {
+        self.socket.shutdown(std::net::Shutdown::Both).unwrap();
+        std::fs::remove_file(&self.address_src).ok();
```
sgkim126: Why do you call `ok`?

basesandbox/src/ipc/semaphore.rs
```diff
@@ -0,0 +1,80 @@
...
+            let semaphore = match mode {
+                Mode::CREATE => libc::sem_open(
+                    name.as_ptr() as *const i8,
+                    nix::fcntl::OFlag::O_CREAT.bits(),
+                    (libc::S_IRWXU | libc::S_IRWXG | libc::S_IRWXO) as libc::c_uint,
+                    0,
+                ),
+                Mode::OPEN => libc::sem_open(name.as_ptr() as *const i8, 0),
+            };
+            if semaphore == libc::SEM_FAILED {
```
sgkim126: Why not
```rust
assert_eq!(semaphore, libc::SEM_FAILED,"Failed to create semaphore: {}", Errno::last())
```

foundry/main.rs
```diff
@@ -21,6 +21,8 @@ extern crate serde_derive;
 #[macro_use]
 extern crate codechain_logger as clogger;
 
+extern crate cbasesandbox;
```
sgkim126: You don't need to add `extern crate`?

basesandbox/src/ipc/semaphore.rs
junha1: I couldn't find any. Any suggestion?

basesandbox/src/execution/executor.rs
sgkim126: I cannot see the protocol which terminates the child process. Isn't it implemented yet?

basesandbox/src/ipc/mod.rs
sgkim126: Do you have a plan to support the socket stream?

basesandbox/src/execution/executor.rs
junha1: @sgkim126 "it to terminate itself" means child just exiting by returning the `main()`. It happens already in this code.

basesandbox/src/ipc/mod.rs
junha1: I'm not sure. If you think such distinction is unnecessary, then I'll just remove it.

junha1:
@sgkim126 Please check

basesandbox/src/test.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    #[test]
+    fn simple_rs() {
+        let mut ctx =
+            executor::execute::<crate::execution::IpcUnixDomainSocket>("./../target/debug/tm_simple_rs", "gmm")
+                .unwrap();
+
+        ctx.ipc.send(b"Hello?\0");
+        let r = ctx.ipc.recv();
+        println!("SERVER recv: {:?}", String::from_utf8(r));
```
sgkim126: You should not use `println` in the test.

coordinator/src/lib.rs
```diff
@@ -48,7 +48,7 @@ pub struct Builder<C: context::Context> {
 }
 
 impl<C: context::Context> Builder<C> {
-    fn new<CTX: context::Context>(ctx: CTX) -> Builder<CTX> {
```
sgkim126: I think clippy said you that `new` function should return `Self` type, doesn't it?
You can make the function to return Self. The `Builder` object already has the type parameter `C`, you don't need to create a new type with a new type parameter.

basesandbox/src/test.rs
sgkim126: What's the purpose of this test?
If this test is to check whether the `recv` method responses, you need to write code like `let _ = ctx.ipc.recv();`, or if you want to check whether the `recv` method returns a specific value, you should write an assertion.
Printing the value in the console does nothing and makes others confused.

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,107 @@
...
+        Executor {
+            child: command.spawn().unwrap(),
+        }
+    }
+}
+
+impl Drop for Executor {
+    fn drop(&mut self) {
+        for _ in 0..10 {
+            std::thread::sleep(std::time::Duration::from_millis(50));
```
sgkim126: If you cannot reason why the child should finish in 500 milliseconds, you should leave the comment.

coordinator/src/lib.rs
junha1: @Byeongjee Please check this comment

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    fn send(&self, data: &[u8]) {
+        self.socket.send_to(data, &self.address_dst).unwrap();
+    }
+
+    fn recv(&mut self) -> Vec<u8> {
+        let (count, address) = self.socket.recv_from(&mut self.buffer).unwrap();
+        address.as_pathname();
+        assert!(count <= self.buffer.len(), "Unix datagram got data larger than the buffer.");
```
sgkim126: I think the send method should have the same assertion.

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,63 @@
...
+
+pub struct DomainSocket {
+    address_src: String,
+    address_dst: String,
+    socket: UnixDatagram,
+    buffer: Vec<u8>,
+}
+
+fn create(address_src: String, address_dst: String) -> DomainSocket {
+    DomainSocket {
```
sgkim126: The current code unnecessarily clones the `address_src`.
Of course, the address_src would not be a big object and cloning might not harm the performance, but it still uses the global heap resource because of the String implementation.
You can rewrite the code if you create a socket first because [`UnixDatagram::bind()`](https://doc.rust-lang.org/std/os/unix/net/struct.UnixDatagram.html#method.bind) takes `AsRef<Path>`.

```rust
    let socket = UnixDatagram::bind(&address_src).unwrap();
    DomainSocket {
        address_src,
        address_dst,
        socket,
        buffer: vec![0; 1024],
    }
```

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    fn send(&self, data: &[u8]) {
+        self.socket.send_to(data, &self.address_dst).unwrap();
+    }
+
+    fn recv(&mut self) -> Vec<u8> {
+        let (count, address) = self.socket.recv_from(&mut self.buffer).unwrap();
+        address.as_pathname();
+        assert!(count <= self.buffer.len(), "Unix datagram got data larger than the buffer.");
+        assert!(
```
sgkim126:
```rust
assert_eq!(address.as_pathname().unwrap(), Path::new(&self.address_dst), "Unix datagram received packet from an unexpected sender.");
```

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+
+pub mod domain_socket;
+pub mod semaphore;
+
+pub trait Ipc: Drop {
```
sgkim126: Should Ipc always implement Drop?

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,63 @@
...
+        create(address_src, address_dst)
+    }
+
+    fn send(&self, data: &[u8]) {
+        self.socket.send_to(data, &self.address_dst).unwrap();
+    }
+
+    fn recv(&mut self) -> Vec<u8> {
+        let (count, address) = self.socket.recv_from(&mut self.buffer).unwrap();
+        address.as_pathname();
```
sgkim126: ?

coordinator/src/lib.rs
Byeongjee: you mean that we should erase the type parameter CTX?
Then I agree with you.

junha1:
@sgkim126 All updated

basesandbox/src/execution/executor.rs
sgkim126: What I meant was not just changing it to `./tmp/Foundry_Socket_Boxer`, but using the temporary directory(`/tmp` ).

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,112 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+
+use crate::ipc::semaphore::{Mode, Semaphore};
+use crate::ipc::Ipc;
+use std::process::Command;
+
+const SOCKET_SRC: &str = "./tmp/Foundry_Socket_Boxer";
+const SOCKET_DST: &str = "./tmp/Foundry_Socket_Boxee";
+const SEMAPHORE: &str = "/Foundry_Semaphore";
```
sgkim126: Do you intend to create a file at the root directory?
Does this code work? You might not create a file because of the permission.

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,112 @@
...
+
+// Note that field's drop is called later that of struct.
+impl<T: Ipc> Drop for Context<T> {
+    fn drop(&mut self) {
+        self.semaphore.wait();
+    }
+}
+
+/// id must be unique for each instance.
+pub fn execute<T: Ipc>(path: &str, id: &str) -> Result<Context<T>, String> {
```
sgkim126: The fixed string id disallows you to run multiple foundry nodes in the same machine.
Change it to u128 and pass the timestamp prefixed random number.

basesandbox/src/execution/executor.rs
junha1: @sgkim126 That doesn't represent an actual path of filesystem.
Read this: https://linux.die.net/man/7/sem_overview

basesandbox/src/execution/executor.rs
junha1: I'll use `H256`. In next PR, I'll add`socket.connect()` not to pass long socket address every time we `send()`.

sgkim126: I found that using a process id and a module id make the name unique, and it doesn't need to hash.
```rust
let pid = std::process::id();
let adddr_src = format!("{}.{}.{}", SOCKET_SRC, pid, id);
```

basesandbox/src/execution/executee.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+pub fn start<T: Ipc>() -> Context<T> {
+    let args: Vec<String> = std::env::args().collect();
+
+    let socket_src_name = &args[1];
+    let socket_dst_name = &args[2];
+    let semaphore_name = &args[3];
+
+    let mut semaphore = Semaphore::new(semaphore_name.to_string(), Mode::OPEN);
```
sgkim126: Is it the only purpose of the semaphore to send the signal that the executor is ready to send the message?
If then, I think you need to move semaphore into the `DomainSocket` implementation. Not all IPC protocols need a semaphore. For instance, a stream socket doesn't need it because it contains such protocol by itself.

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,64 @@
...
+// You should have received a copy of the GNU Affero General Public License
+
+use super::Ipc;
+use std::os::unix::net::UnixDatagram;
+use std::path::Path;
+
+pub struct DomainSocket {
+    address_src: String,
+    address_dst: String,
+    socket: UnixDatagram,
```
sgkim126: Why did you use a datagram instead of a stream socket?
If you used a stream socket, you don't need to implement semaphore.

junha1: @sgkim126 Because it's easier. Do you want me to rewrite this PR with stream?

junha1:
This work will be updated, and will be included in another PR

PR #250, opened 2020-03-11
Send and relay packet

majecty:
I wrote code to send and relay packets.
~While I testing it, I met an unexpected error: "Illformed Sequence stored in the DB: RlpExpectedToBeData".
I'll fix it soon.~

### DIFF
--- a/core/src/ibc/channel_04/log.rs
+++ b/core/src/ibc/channel_04/log.rs
@@ -38,7 +38,6 @@ pub fn set_packet<'a>(
     }
 }
 
-#[allow(dead_code)]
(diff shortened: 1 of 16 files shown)


core/src/ibc/channel_04/log.rs
```diff
@@ -38,7 +38,6 @@ pub fn set_packet<'a>(
     }
```
junha1: Typo in the commit message 'remove_pcket'

majecty: I fixed it. Thanks!

PR #255, opened 2020-03-12
Assign unbound port numbers in test instances

### LINKED ISSUE #254: Assign unbound port numbers in e2e tests
Currently, e2e tests run Foundry instances with predefined port numbers. (ex. `8080 + id`)
This is problematic if the port is already in use.
We should find and assign an unbound port number to each Foundry instance.
I suspect that this is the cause of some e2e test failures.

Byeongjee:
Resolves #254 

Port numbers assigned to test Foundry instances should be unbound.
If the port is already in use, the test will fail, which is not an intended behavior.


test/src/helper/spawn.ts
```diff
@@ -160,6 +163,7 @@ export default class CodeChain {
         const { chain, argv, additionalKeysPath, env } = options;
         this._id = CodeChain.idCounter++;
 
+        this._port = 3486 + this.id;
         const { rpcPort = 8081 + this.id } = options;
         this._rpcPort = rpcPort;
```
majecty: How about not initializing port and rpcPort in the consturctor?
The port numbers initialized here are not used. These values will make a user confuse.

Byeongjee: I was fixing some minor changes. Sorry for the inconvenience.

Byeongjee: port and rpcPort cannot be initialized in the constructor, because `getPort` is async so it cannot be called in the constructor.

Byeongjee: The values initialized there indicate the preference, and they will be used in the most time.
`initialize_port` will change the port into an unbound one only if it's already in use.

majecty: Although the values are preferenced values, they should not be accessible in the interface. Since they can be accessible by `get port()` and `get rpcPort()` it can be confusing. If you want to set the preferences in the constructor function, please create preferredRPCPort and preferredPort fields.

Byeongjee: I misread what you commented. sorry. You are right. I'll remove it.

PR #256, opened 2020-03-12
Remove omitted field from ImportRoute

### LINKED ISSUE #230: Remove ImportRoute.omitted
We are not using a PoW consensus engine. Since the `omitted` field in `ImportRoute` is reserved only for PoW engines, it should be removed. (https://www.notion.so/codechain/Enacted-VS-Imported-VS-Omitted-ea5d04ae319d4da38f9c91aa940d49c5)

If we remove the `omitted` field, we don't need the `ImportRoute` struct anymore. Maybe we can change the return type of `BlockChain::insert_block` from `ImportRoute` to `bool`. In this case, `true` will indicate that the block is successfully inserted to the chain, and false will indicate that the block is already inserted.

Byeongjee:
Resolves #230 

We don't use the `omitted` field anymore.

I'm not sure if I should remove `ImportRoute` struct at all, or leave it. 
Please give a comment.

### DIFF
--- a/core/src/blockchain/update_result.rs
+++ b/core/src/blockchain/update_result.rs
@@ -1,0 +1,69 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use super::block_info::{BestBlockChanged, BestHeaderChanged};
+use ctypes::BlockHash;
+
+#[derive(Debug, PartialEq)]
+pub struct ChainUpdateResult {
+    // Some(updated_best_block_hash) if chain is updated
+    // None if chain is not updated
+    pub enacted: Option<BlockHash>,
+}
+
+impl ChainUpdateResult {
+    pub fn new(best_block_changed: &BestBlockChanged) -> Self {
+        match best_block_changed {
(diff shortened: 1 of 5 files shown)


test/package.json
```diff
@@ -25,28 +25,28 @@
...
     "prettier": "^1.19.1",
-    "ts-node": "^8.5.2",
     "ts-jest": "^23.1.3",
+    "ts-node": "^8.5.2",
     "tslint": "^5.20.1",
     "tslint-config-prettier": "^1.18.0",
-    "typedoc": "^0.14.2",
     "tslint-no-unused-expression-chai": "^0.1.4",
+    "typedoc": "^0.16.11",
     "typescript": "^3.7.2"
```
majecty: Please remove unrelated changes. Or you can create a commit for this.

Byeongjee: Sorry for making mistakes often.  This was not intended. I'll remove it.

Byeongjee:
A problem of `ImportResult` is that it's already in use. https://github.com/CodeChain-io/foundry/blob/d8d4e7c0d18ff9ddf0418452b6a34c8ebafab1cb/core/src/client/mod.rs#L260

majecty:
@Byeongjee How about `EnactedResult`?

Byeongjee:
I was thinking about `InsertResult` or `ChainUpdateResult`.
`EnactResult` would be fine as well.

Byeongjee:
Btw I'll move this struct into `blockchain` module.

majecty:
I like `ChainUpdateResult`!

Byeongjee:
I'm not sure where to relocate the `ChainUpdateResult`. Do you have a suggestion?

majecty:
Using blockchain/mod.rs or Creating a new file in the src/blockchain/ directory seems good.

Byeongjee:
I created `update_result` submodule.

core/src/blockchain/update_result.rs
```diff
@@ -0,0 +1,68 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use super::block_info::{BestBlockChanged, BestHeaderChanged};
+use ctypes::BlockHash;
+
+#[derive(Debug, PartialEq)]
+pub struct ChainUpdateResult {
+    /// Blocks that were validated by new block.
```
majecty: Please update the comment.

PR #257, opened 2020-03-12
Modify logic related to the Tx type for checking whether a Tx is verified or not

### LINKED ISSUE #168: Unify UnverifiedTransaction and SignedTransaction
Previously in `ECDSA` signing algorithm, it is possible to recover a public key from a signature. Since the `Ed25519` is newly adopted, we could not utilize the recovery function.
A `SignedTransaction` was constructed from an `UnverifiedTransaction` when the signature recovery returned a public key without errors. We don't need such an indirection because `UnverifiedTransaction` should include a public key.

somniumism:
Summary: We modified logic related to the transaction type for checking whether a transaction is verified or not. From now on, there are two transaction types, `UnverifiedTransaction` and `VerifiedTransaction`, and the `UnverifiedTransaction` can be converted to `VerifiedTransaction` using the method `try_into()`. During this conversion, the signature is verified through the method `verify_transaction()`.

---

We were using two transaction types, `UnverifiedTransaction` and `SignedTransaction`. The `SignedTransaction` was constructed from an `UnverifiedTransaction` when the signature recovery returned a public key without errors. These structs were as follows:
```rust
pub struct UnverifiedTransaction {
    /// Plain Transaction.
    unsigned: Transaction,
    /// Signature.
    sig: Signature,
    /// Signer's public key
    signer_public: Public,
    /// Hash of the transaction
    hash: TxHash,
}

pub struct SignedTransaction {
    tx: UnverifiedTransaction,
    signer_public: Public,
}
```

We don't need such an indirection because `UnverifiedTransaction` should include a public key. Then, we had thought it would be good to unify `SignedTransaction` and `UnverifiedTransaction`. However, we still need logic to distinguish whether each transaction is verified or not. 

Therefore, we modified the logic we were used(`UnverifiedTransaction` and `SignedTransaction`) to `UnverifiedTransaction` and `VerifiedTransaction`. These struct have the same components and we can check whether it is a verified transaction or not. These structs are as follows:
```rust
pub struct SignedTransaction {
    /// Plain Transaction.
    unsigned: Transaction,
    /// Signature.
    sig: Signature,
    /// Signer's public key
    signer_public: Public,
    /// Hash of the transaction
    hash: TxHash,
}

pub struct VerifiedTransaction(SignedTransaction);

pub struct UnverifiedTransaction(pub SignedTransaction);
```

From now on, we can verify transactions in the process of converting from `UnverifiedTransaction` to `VerifiedTransaction`. The trait `TryFrom` is implemented for `VerifiedTransaction` and we can convert the unverified transaction to it using a method `.try_into()`. During this conversion, the signature is verified the method `verify_transaction()`.

### DIFF
--- a/core/src/miner/backup.rs
+++ b/core/src/miner/backup.rs
@@ -16,10 +16,45 @@
 
 use super::mem_pool_types::MemPoolItem;
 use crate::db as dblib;
+use crate::error::Error;
+use crate::miner::mem_pool_types::TxOrigin;
+use crate::transaction::UnverifiedTransaction;
+use ctypes::BlockNumber;
 use kvdb::{DBTransaction, KeyValueDB};
 use primitives::H256;
 use rlp::Encodable;
 use std::collections::HashMap;
+use std::convert::{TryFrom, TryInto};
+
+pub type PoolingInstant = BlockNumber;
+
+#[derive(Clone, Eq, PartialEq, Debug, RlpEncodable, RlpDecodable)]
+pub struct MemPoolItemProjection {
+    /// Transaction.
+    pub tx: UnverifiedTransaction,
+    /// Transaction origin.
+    pub origin: TxOrigin,
+    /// Insertion time
+    pub inserted_block_number: PoolingInstant,
+    /// Insertion timstamp
+    pub inserted_timestamp: u64,
+    /// ID assigned upon insertion, should be unique.
+    pub insertion_id: u64,
+}
+
+impl TryFrom<MemPoolItemProjection> for MemPoolItem {
+    type Error = Error;
+    fn try_from(mem_pool: MemPoolItemProjection) -> Result<Self, Error> {
+        let verified_tx = mem_pool.tx.try_into()?;
+        Ok(MemPoolItem {
+            tx: verified_tx,
+            origin: mem_pool.origin,
+            inserted_block_number: mem_pool.inserted_block_number,
+            inserted_timestamp: mem_pool.inserted_timestamp,
+            insertion_id: mem_pool.insertion_id,
+        })
+    }
+}
 
 const PREFIX_SIZE: usize = 5;
 const PREFIX_ITEM: &[u8; PREFIX_SIZE] = b"item_";
@@ -46,9 +81,9 @@ pub fn recover_to_data(db: &dyn KeyValueDB) -> HashMap<H256, MemPoolItem> {
     for (key, value) in db.iter(dblib::COL_MEMPOOL) {
         let bytes = (*value).to_vec();
         let rlp = rlp::Rlp::new(&bytes);
-
         let decoded_key = (key.as_ref()[PREFIX_SIZE..]).into();
-        let decoded_item = rlp.as_val().unwrap();
+        let mem_pool_projection: MemPoolItemProjection = rlp.as_val().unwrap();
+        let decoded_item = mem_pool_projection.try_into().expect("DB corruption detected");
         by_hash.insert(decoded_key, decoded_item);
     }
 
--- a/core/src/miner/mem_pool.rs
+++ b/core/src/miner/mem_pool.rs
@@ -22,7 +22,7 @@ use super::mem_pool_types::{
 use super::TransactionImportResult;
 use crate::client::{AccountData, BlockChainTrait};
 use crate::miner::fetch_account_creator;
-use crate::transaction::{PendingSignedTransactions, SignedTransaction};
+use crate::transaction::{PendingVerifiedTransactions, VerifiedTransaction};
 use crate::Error as CoreError;
 use ckey::{public_to_address, Ed25519Public as Public};
 use ctypes::errors::{HistoryError, RuntimeError, SyntaxError};
@@ -246,7 +246,7 @@ impl MemPool {
         for input in inputs {
             let tx = input.transaction;
             let signer_public = tx.signer_public();
-            let seq = tx.seq;
+            let seq = tx.transaction().seq;
             let hash = tx.hash();
 
             let origin = if input.origin.is_local() && !self.is_local_account.contains(&signer_public) {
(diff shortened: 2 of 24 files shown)


somniumism:
@HoOngEe I think it would be better to squash the commits(`Replace ~~`) into one commit after your review. Before you approve this PR, please mention me. : )

HoOngEe:
I think it's better to unify the two types into 
```
enum SignedTransaction {
  Unverified(Transaction),
  Verified(Transaction),
};
```
Previous constructions of `SignedTransaction` can be replaced with variant change from `Unverified` into `Verified` invoking `verify()` of `Ed25519`.

core/src/transaction.rs
```diff
@@ -74,53 +67,146 @@ impl rlp::Decodable for UnverifiedTransaction {
...
-impl UnverifiedTransaction {
+impl From<Verified> for Unverified {
+    fn from(tx: Verified) -> Self {
+        Unverified(tx.0)
+    }
+}
+
+impl Verified {
+    /// Try to verify transaction and recover public.
+    pub fn try_new(tx: Unverified) -> Result<Self, Error> {
```
HoOngEe: Let's remove this method which converts `UnverifiedTransaction` into `VerifiedTransaction` without verification, and utilize `try_into()` instead.

core/src/transaction.rs
```diff
@@ -74,53 +67,146 @@ impl rlp::Decodable for UnverifiedTransaction {
...
+pub struct Unverified(pub TempTransaction);
+
+impl rlp::Encodable for Verified {
     fn rlp_append(&self, s: &mut RlpStream) {
-        self.rlp_append_sealed_transaction(s)
+        self.0.rlp_append(s);
+    }
+}
+
+impl rlp::Decodable for Verified {
```
HoOngEe: This path may be violate the invariant that `VerifiedTransaction` was always verified. If no usage is found then remove it or convert it first to `Unverified` then utilize `try_into`

core/src/transaction.rs
```diff
@@ -74,53 +67,146 @@ impl rlp::Decodable for UnverifiedTransaction {
...
+        s.append(&self.unsigned.fee);
+        s.append(&self.unsigned.network_id);
+        s.append(&self.unsigned.action);
+        s.append(&self.sig);
+        s.append(&self.signer_public);
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Verified(TempTransaction);
```
HoOngEe: Let's rename `Verified` to `VerifiedTransaction` and `Unverified` to `UnverifiedTransaction`.

somniumism: I renamed it!

somniumism:
I couldn't remove `rlp::Decodable` because it is used by `MemPoolItem`

HoOngEe:
The `MemPoolItem` utilizes `Decode` when mempool is recovered from db in case of shut-down. I think it's better to remove the `Decode` from `VerifiedTransaction` which can violate the invariant of the type. You can implement `Decode` for `MemPoolItem` using `Decode` for `UnverifiedTransaction` and `try_into()`.

core/src/transaction.rs
```diff
@@ -74,53 +67,147 @@ impl rlp::Decodable for UnverifiedTransaction {
...
+
+impl From<VerifiedTransaction> for UnverifiedTransaction {
+    fn from(tx: VerifiedTransaction) -> Self {
+        UnverifiedTransaction(tx.0)
+    }
+}
+
+impl VerifiedTransaction {
+    /// Try to verify transaction and recover public.
+    pub fn try_new(tx: UnverifiedTransaction) -> Result<Self, Error> {
```
HoOngEe: Let's remove this method and replace all the `try_new()`s with `try_into()`s

core/src/transaction.rs
```diff
@@ -168,12 +159,12 @@ impl Deref for UnverifiedTransaction {
...
 impl TryFrom<UnverifiedTransaction> for VerifiedTransaction {
-    type Error = &'static str;
-    fn try_from(tx: UnverifiedTransaction) -> Result<Self, Self::Error> {
+    type Error = Error;
+    fn try_from(tx: UnverifiedTransaction) -> Result<Self, Error> {
         if tx.verify_transaction() {
             Ok(VerifiedTransaction(tx.0))
         } else {
-            Err("")
+            Err(Error::Transaction(TransactionError::InvalidVerification))
```
HoOngEe: I think `InvalidVerification` implies the "verification" itself is invalid. What about "InvalidSignature"?

somniumism: Good idea! I modified it!

somniumism:
@HoOngEe I added a commit to resolve the test failures in the e2e test. Could you check [the commit](https://github.com/CodeChain-io/foundry/pull/257/commits/45792ad57fa4af10bfa23f3531bf382f6d447296)?

core/src/error.rs
```diff
@@ -126,6 +126,20 @@ impl fmt::Display for SchemeError {
     }
 }
 
+#[derive(Debug, PartialEq, Clone, Copy, Eq)]
+pub enum TransactionError {
```
HoOngEe: For the `Error` part, we already have `KeyError::InvalidSignature`. Let's use it instead.

somniumism: I removed `TransactionError`. Thank you!

core/src/transaction.rs
```diff
@@ -74,141 +66,158 @@ impl rlp::Decodable for UnverifiedTransaction {
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct UnverifiedTransaction(SignedTransaction);
+
+impl rlp::Encodable for VerifiedTransaction {
+    fn rlp_append(&self, s: &mut RlpStream) {
+        self.0.rlp_append(s);
+    }
+}
+
+impl rlp::Decodable for VerifiedTransaction {
```
HoOngEe: I'm sorry that I recommended you to implement `rlp::Decodable` for `VerifiedTransaction` for the case of db backup. However it is inappropriate. `decode` is just in charge of converting serialized bytes into target structs. In case of db corruption, it seems ok to be panicked when recovering `MemPool` from the db. However, this general method may be used somewhere else then the `expect` and `panic` seem dangerous. I think it's better to elevate the panic behavior to higher level such as `recover_to_data` function.

core/src/transaction.rs
```diff
@@ -74,141 +66,158 @@ impl rlp::Decodable for UnverifiedTransaction {
...
+    }
+}
+
+impl VerifiedTransaction {
+    /// Signs the transaction as coming from `signer`.
+    pub fn new_with_sign(tx: Transaction, private: &Private) -> VerifiedTransaction {
+        let sig = sign(&tx.hash(), private);
+        UnverifiedTransaction::new(tx, sig, private.public_key())
+            .try_into()
+            .expect("secret is valid so it's recoverable")
```
HoOngEe: Let's modify the comment because we're not using the previous key recovery

somniumism: I modified this comment to `The transaction's signature is invalid". How about this?

core/src/transaction.rs
somniumism: I modified it. Could you check the last commit?

core/src/miner/backup.rs
```diff
@@ -16,10 +16,51 @@
...
+    pub inserted_timestamp: u64,
+    /// ID assigned upon insertion, should be unique.
+    pub insertion_id: u64,
+}
+
+impl TryFrom<MemPoolItemProjection> for MemPoolItem {
+    type Error = Error;
+    fn try_from(mem_pool: MemPoolItemProjection) -> Result<Self, Error> {
+        let tx = mem_pool.tx;
+        if tx.verify_transaction() {
```
HoOngEe: Do we need to double check the content of the `UnverifiedTransaction`? tx.verify_transaction() will be called inside of the `tx.try_into()`. Am I missing something?

core/src/miner/backup.rs
```diff
@@ -46,9 +87,9 @@ pub fn recover_to_data(db: &dyn KeyValueDB) -> HashMap<H256, MemPoolItem> {
     for (key, value) in db.iter(dblib::COL_MEMPOOL) {
         let bytes = (*value).to_vec();
         let rlp = rlp::Rlp::new(&bytes);
-
         let decoded_key = (key.as_ref()[PREFIX_SIZE..]).into();
-        let decoded_item = rlp.as_val().unwrap();
+        let mem_pool_projection: MemPoolItemProjection = rlp.as_val().unwrap();
+        let decoded_item = mem_pool_projection.try_into().expect("DB corruption failed");
```
HoOngEe: "DB corruption detected" seems more appropriate.

core/src/miner/backup.rs
HoOngEe: What do you think about changing `verify_transaction` method to be `private`? Only `try_into` is enough to be public because verification will only be conducted to convert unverified one into verified one.

core/src/miner/mem_pool.rs
```diff
@@ -983,42 +983,6 @@ pub mod test {
         rlp_encode_and_decode_test!(TxOrigin::External);
     }
 
-    #[test]
```
HoOngEe: What about testing the conversion path: `VerifiedTransaction` -> `Bytes` -> `UnverifiedTransaction` -> `VerifiedTransaction` ?

somniumism: I modified the tests.

core/src/miner/backup.rs
somniumism: I fixed it.

core/src/miner/backup.rs
somniumism: If we change `verified_transaction()` to `private method`, we can use it on `VerfiedTransaction`'s `Impl` because `verified_transaction()` is implemented for `UnverifiedTransaction`. So, I didn't modified it.
However, double-check is my mistake, so I modified it. Thank you for a good review.

PR #260, opened 2020-03-13
Implement validator interface call

Byeongjee:
- [x] Fix the mempool code to use validator interface
- [x] Implement block preparation logic (WIP. need to fix close logic)
- [x] Implement block verification logic 
- [x] Implement `commit` and `revert` interface
- [x] Implement evidence handling logic
- [x] Implement chain initialization logic
- [x] Remove application-level code from `core`
    - [x] Remove account from miner
    - [x] Remove account from client
    - [x] Remove account from engine

I finished implementing evidence handling and block execution logic.

Remaining tasks

- [x] Add transaction signature verification logic
- [x] Remove txes from mempool when the block is accepted (easy)
- [x] Make `validator::Evidence` more informative
- [x] Add evidence hash to block header
- ~~[ ] Add logic to remove old evidenecs~~ -> It seems that we don't need this in the MVP
- [x] Update validator set after each block execution (can do inside `block.rs`)
- [x] Update common params after each block execution (can do inside `block.rs`)
- [x] Simplify and rename `AccountProvider` usedf by `EngineSigner`
- [x] `initialize_chain`


Byeongjee:
Block preparation logic will be refactored after merging https://github.com/CodeChain-io/foundry/pull/264

Byeongjee:
@sgkim126 Could you review this?
It's still in progress, so you don't have to look at all changes, but I would be happy if you check these points.

1. I removed `state` from Block, because state transition will be done by `execute_block` call.
2. I removed codechain_machine module, because I thought that it is responsible for doing application-level jobs.
	1. I’ll move params field into engine. Until then, params are hardcoded
3. I changed block importing logic
	1. Previously, block execution and block importing was strictly decoupled (because blocks had states]
	2. But now it's not. Now I have functions like `execute_and_check_block`, and `prepare_and_execute_block`.
	3. It seems slightly weird to me, but I’m not sure how to improve this

Byeongjee:
Please have a look at block importing/preparation/verification logic.
They are my main concerns.

Byeongjee:
I moved `genesis_common_params` into engine.

sgkim126:
> @sgkim126 Could you review this?
> It's still in progress, so you don't have to look at all changes, but I would be happy if you check these points.
> 
> 1. I removed `state` from Block, because state transition will be done by `execute_block` call.

I cannot agree on [this commit](https://github.com/CodeChain-io/foundry/pull/260/commits/777f7ab10d5d198bf42d11c9c73e3379bca69179). The commit makes a type with an implicit internal state. It merely increases the probability of bugs rather than making code simple. Please revive `OpenBlock` and `ClosedBlock`.
Because you need to distinguish `OpenBlock` and `CloseBlock` types, it would be nice not to erase `ExecutedBlock` to reduce duplicated code.


> 2. I removed codechain_machine module, because I thought that it is responsible for doing application-level jobs.
>
>    1. I’ll move genesis_common_params field into engine or somewhere else. Until then, params are hardcoded

I agree to remove CodeChainMachine. To implement it should be the role of the modules, and the application should have interfaces to do that. However, you should clean-up [the commit](https://github.com/CodeChain-io/foundry/pull/260/commits/cff2beddf5d64da8627a9dec0af0fac2fb04398e). The commit does two things. It removes CodeChainMachine and CodeChangeEngine. You'd better split the commit into two.


> 3. I changed block importing logic
>    
>    1. Previously, block execution and block importing was strictly decoupled (because blocks had states]
>    2. But now it's not. Now I have functions like `execute_and_check_block`, and `prepare_and_execute_block`.
>    3. It seems slightly weird to me, but I’m not sure how to improve this

I cannot understand the reason you want to change executing and importing to `execute_and_check_block` and `prepare_and_executed_block`. You will still need to separate importing and block execution.

Byeongjee:
@sgkim126 If a block has an internal state, how can the host update the state in the block?
I removed `state` from `Block` because I didn't want `execute_block` to pass `state` as an argument or as a return value.

Byeongjee:
@sgkim126 If a block has an internal state, how should the host update the state?
I thought that passing `state` as an argument or a return value of `execute_block` was not a nice idea.

sgkim126:
> @sgkim126 If a block has an internal state, how should the host update the state?
> I thought that passing `state` as an argument or a return value of `execute_block` was not a nice idea.

What is a state that you mean? The executed result of the transactions in a block is a state. It's natural that block has a state.

Byeongjee:
What I meant by `state` is the state of the application, or the state of the latest best block.

To call `execute_block`, the host has to provide the application about the previous state. 
Also, after `execute_block` the host has to fetch the application state after the execution.

I thought we shouldn't expose `TopLevelState` to the application. That was my point.

Byeongjee:
Simply speaking, I think `state` should not be on the host side. That is my main concern.
The host can see `state`, as a key-value DB, but it shouldn't do anything more.

IMO state should belong strictly to the application, and all the state transitions should be done only by the ABCI.

I'm not sure how to do that, with `state` in blocks.

Byeongjee:
Btw I separated commits 
- Remove CodeChainEngine
- Remove codechain_machine module

sgkim126:
Okay. Now I got it. You can move the state(precisely speaking `TopLevelState`) to the application side, but you still need the state in the host side. The host needs a state root and a logs root of the block that is executing and, transactions included in the block.
Executing transactions, sealing a block and propagating and validating a block is still roles of the host.

Byeongjee:
@sgkim126 I was misunderstanding something. I want to clarify some points that I got from a discussion with @HoOngEe 

Key-value DB management and trie management will be done by the host.
What the application sees is a portion of key-value DB, which corresponds to some specific state.

From this point of view, I think it makes sense to have code like `state.execute_block()`.
I think it's up to our design decision. Please correct me if I'm misunderstanding.

Byeongjee:
SImply speaking, the host manages `Trie` (as a key-value DB) and the application is responsible for executing transactions based on the given `Trie`, right?

Byeongjee:
I finished implementing evidence handling and block execution logic.

Remaining tasks

- [ ] Remove txes from mempool when the block is accepted (easy)
- [ ] Make `validator::Evidence` more informative
- [ ] Add evidence hash to block header
- [ ] Load common params from state? (need discussion)
- [ ] `initialize_chain`
- [ ] Simplify and rename `AccountProvider` usedf by `EngineSigner`

Byeongjee:
@sgkim126 Could you review this WIP PR again?

- I changed the block execution interface as we discussed.
- I changed InvoiceDB to EventDB
- I added Evidence to block and implemented evidence handling logic (`EvidenceCollector`)
     - Content of evidence is not yet decided

Byeongjee:
I'm figuring out how to avoid unncecesary copy of `event` and `evidence`.

core/src/miner/miner.rs
```diff
@@ -486,17 +264,22 @@ impl MinerService for Miner {
...
         C: AccountData + BlockChainTrait + BlockProducer + EngineInfo + ImportBlock, {
         ctrace!(MINER, "chain_new_blocks");
 
         {
-            let fetch_account = fetch_account_creator(chain);
-            let current_block_number = chain.chain_info().best_block_number;
-            let current_timestamp = chain.chain_info().best_block_timestamp;
             let mut mem_pool = self.mem_pool.write();
-            mem_pool.remove_old(&fetch_account, current_block_number, current_timestamp);
+            enacted
```
Byeongjee: @sgkim126 Could you check this, too?
This code removes enacted transactions from the mempool.
I'm worried if this is wrong, or too inefficient.

sgkim126: It seems okay to me.
I can't think of a more efficient way now and, at least, the implementation is not wrong.

core/src/client/test_client.rs
```diff
@@ -192,30 +188,18 @@ impl TestBlockChainClient {
         }
     }
     /// Add a block to test client with designated author.
-    pub fn add_block_with_author(&self, author: Option<Address>, n: usize, transaction_length: usize) -> BlockHash {
+    pub fn add_block_with_author(&self, author: Option<Public>, n: usize, transaction_length: usize) -> BlockHash {
```
sgkim126: Do we need to change `Address` to `Public`?

core/src/block.rs
```diff
@@ -16,7 +16,7 @@
 
 use super::invoice::Invoice;
 use crate::client::{EngineInfo, TermInfo};
-use crate::consensus::CodeChainEngine;
```
sgkim126: Please fix the typo in the commit message.
`CodeChainEngineCompared with ConsensusEngine` -> `CodeChainEngine compared to ConsensusEngine`

core/src/client/test_client.rs
```diff
@@ -531,13 +531,11 @@ impl super::EngineClient for TestBlockChainClient {
 
 impl EngineInfo for TestBlockChainClient {
     fn network_id(&self) -> NetworkId {
-        // need to move common params from machine to something else
```
sgkim126: I think it's better to merge this commit with the [parent](https://github.com/CodeChain-io/foundry/pull/260/commits/500294b1f41b5f3970d4eff75868904424e73fae).

core/src/block.rs
```diff
@@ -146,59 +149,40 @@ impl<'x> OpenBlock<'x> {
...
-        tx: VerifiedTransaction,
-        h: Option<TxHash>,
-        client: &C,
-        parent_block_number: BlockNumber,
-        parent_block_timestamp: u64,
-    ) -> Result<(), Error> {
-        if self.block.transactions_set.contains(&tx.hash()) {
-            return Err(HistoryError::TransactionAlreadyImported.into())
-        }
+    pub fn begin_transaction_execution(&mut self, coordinator: &Coordinator, evidences: Vec<Evidence>) {
```
sgkim126: Why is it named `begin_transaction_execution` instead of `open_block`?
I think the current name can confuse users to believe it should be called before every `execute_transactions`.

core/src/client/test_client.rs
Byeongjee: A validator in the host is represented as a tuple of `Public` and `VotePower = u64`and no more information is exposed to the host.
Therefore there is no reason to keep the concept of `Address` in the host because `Address` is nothing more than a hash of `Public` (It doesn't imply anything about account, etc)
But you are right, we don't really have to change it.

Byeongjee: I'll not change it now. Let's talk about it later.

core/src/block.rs
Byeongjee: I fixed it

core/src/client/test_client.rs
Byeongjee: Ok

core/src/block.rs
Byeongjee: I thought `open_block.open_block()` was an confusing naming.
I agree that the current name is also confusing as well.
Could you suggest any better name?
Or I can merge the method into `OpenBlock::try_new`, but in that case, the `try_new` method will become complicated.

Byeongjee:
These tasks should be done after implementing each module
- [ ] Remove unused functionality from AccountProvider in EngineSigner
- [ ] Remove transaction types
- [ ] Remove codechain_machine
- [ ] Add more host-side RPCs
- [ ] Remove auto self nominate
- [ ] Remove commonParams
- [ ] Remove account /action_data from state and fix initialize_state

Byeongjee:
In this PR, I decoupled application level concepts from the host level concepts (in order to use `Validator` interface)
- I didn’t remove the application level concepts yet unless it was hard to fix compilation issues without removing them.
- Remaining application-level concepts should be removed by module implementing PRs

Byeongjee:
@sgkim126 Major Changes after your last review
- I cleaned up and implemented unit tests
- I tried to add more error types when txes are rejected by mempool (or miner)
- I create `ConsensusParams`. `CommonParams` will be removed
- I create `NextValidatorSet` and `CurrentValidatorSet` layer
	- This provides host-side interface for accessing validator set
	- `NextValidators` and `CurrentValidators` will be moved to Staking module.
	- After then,  the implementation of  `NextValidatorSet` and `CurrentValidatorSet` will be changed
- I implemented `close_block` related logic
	- Validator set and consensus params will be updated through `close_block`
		- `close_block` will return next validator set and next consensus params
- I fixed foundry initialization logic 
	- I changed AccountProvider initialization logic (in Tendermint Worker)
	- Now I use `initialize_chain` interface
- I fixed Transaction type in `Validator` interface in order to avoid unnecessary copies.
- I removed some RPCs, some types and traits because it was the easiest way to compile my code
	- Anyway they are planned to be removed after implementing modules, …e

Please forgive that some commits are not organized nicely (there are some small commits which are better to be merged to past commits)
This is because the first 20~30 commits are not compiling, and it was quite costly to modify the not compiling commits. Tell me if there are not tolerable commits.

Byeongjee:
@sgkim126 I found out that I forgot to implement logic to remove old evidences.
I'll implement it after we decide how to handle `term`.

state/src/impls/test_helper.rs
```diff
@@ -16,8 +16,10 @@
 
 #![macro_use]
 
+#[allow(unused)]
```
sgkim126, line 19: impolemented -> implemented
in the commit message

coordinator/src/validator.rs
```diff
@@ -213,7 +213,11 @@ impl Decodable for TransactionWithGas {
 }
 
 pub enum Evidence {
-    DoubleVote, // Should import and use DoubleVote type defined in tendermint module?
+    DoubleVote {
```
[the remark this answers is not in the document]
Byeongjee: I wanted not to pass the signature to the application.
Since evidence verification is done by the host, the application does not need information to verify the evidence. 
It only needs information to penalize validators.
This is what Tendermint ABCI does.

Do you think this behavior is problematic? Or are you not satisfied with the name Evidence?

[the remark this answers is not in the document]
Byeongjee: If you are not satisfied with the name, I can rename it to something like`VerifiedCrime`

coordinator/src/validator.rs
```diff
@@ -11,11 +11,24 @@ pub type Bytes = Vec<u8>;
 
 pub type VoteWeight = u64;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Event {
     pub key: &'static str,
     pub value: Bytes,
 }
 
+impl Encodable for Event {
```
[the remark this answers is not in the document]
Byeongjee: Encoded, Decoded traits are required because I need to store Events to eventDB.
I think it was a mistake to put the change in this commit.
I'll create a separate commit before the commit implementing EventDB.
Please tell me if you think I should squash the commits.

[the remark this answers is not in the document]
sgkim126: I think it's your choice. If you ask me, I'll separate the commits when the commit implementing EventDB is too big, otherwise, I'll squash them into one.

state/src/impls/test_helper.rs
Byeongjee: Fixed

core/src/miner/mem_pool_types.rs
```diff
@@ -14,362 +14,56 @@
...
-    /// Memory usage of the external transactions in the queue
+    pub pool: HashMap<TxHash, TransactionWithMetadata>,
+    /// Memory usage of the transactions in the queue
     pub mem_usage: usize,
     /// Count of the external transactions in the queue
     pub count: usize,
 }
 
-impl CurrentQueue {
+impl<'a> TransactionPool {
```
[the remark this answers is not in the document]
Byeongjee: Thanks.
I don't understand why Clippy is not complaining about this?
Does adding `<'a>` change the behavior of code?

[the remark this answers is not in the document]
Byeongjee: Anway fixed.

core/src/client/client.rs
```diff
@@ -318,18 +318,6 @@ impl EngineInfo for Client {
                 .seq()
         })
     }
-
-    fn possible_authors(&self, block_number: Option<u64>) -> Result<Option<Vec<PlatformAddress>>, EngineError> {
```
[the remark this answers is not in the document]
Byeongjee: I just realized that not only the commit message, but the commit itself was a mistake.
SInce `possible_authors` just returns a validator set, we don't have to remove this.
I'll restore this.
Maybe we should change the type of `PlatformAddress` in the future.

coordinator/src/validator.rs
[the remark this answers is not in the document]
Byeongjee: Since the name `Evidence` is defined a long time ago, I'll just create a new commit to rename it.

coordinator/src/lib.rs
```diff
@@ -40,9 +40,10 @@ impl validator::Validator for Coordinator {
     }
 
     fn remove_transactions(
-        &mut self,
```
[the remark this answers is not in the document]
Byeongjee: I think lifetime parameters are defined correctly in the final commit.
I made a mistake in some old commits.
I'll fix them.

coordinator/src/validator.rs
```diff
@@ -143,6 +199,19 @@ impl<'a> TransactionWithGas {
         self.tx.hash()
     }
 }
+
+impl Encodable for TransactionWithGas {
```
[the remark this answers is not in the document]
Byeongjee: I mistakenly added this and removed this later. It is not used at all. I'll remove this.

coordinator/src/validator.rs
```diff
@@ -122,6 +166,18 @@ impl<'a> TransactionWithMetadata {
     }
 }
 
+impl Encodable for TransactionWithMetadata {
```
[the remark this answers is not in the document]
Byeongjee: I'll move this to the commit implementing new mempool data types.
In the commit, we make mempool to backup TransactionWithMetadata.
That's why we need Encodable for TransactionWithMetadata.

coordinator/src/validator.rs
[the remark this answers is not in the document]
Byeongjee: I created a separate commit before the commit implementing EventDB.

coordinator/src/validator.rs
```diff
@@ -78,6 +109,18 @@ impl Transaction {
     }
 }
 
+impl Encodable for Transaction {
```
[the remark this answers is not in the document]
Byeongjee: I create a separate commit before the commit storing these tx types into DB.

coordinator/src/lib.rs
```diff
@@ -36,7 +36,7 @@ impl validator::Validator for Coordinator {
         unimplemented!()
     }
 
-    fn fetch_transactions_for_block(&self, transactions: Vec<&TransactionWithMetadata>) -> Vec<&TransactionWithGas> {
+    fn fetch_transactions_for_block(&self, transactions: Vec<&TransactionWithMetadata>) -> Vec<TransactionWithGas> {
```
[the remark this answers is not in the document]
Byeongjee: Fixed.

coordinator/src/lib.rs
```diff
@@ -36,16 +36,19 @@ impl validator::Validator for Coordinator {
         unimplemented!()
     }
 
-    fn fetch_transactions_for_block(&self, transactions: Vec<&TransactionWithMetadata>) -> Vec<TransactionWithGas> {
+    fn fetch_transactions_for_block<'a>(
```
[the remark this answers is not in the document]
Byeongjee: I organized the commits that changes the validator interface.

coordinator/src/validator.rs
```diff
@@ -227,12 +227,18 @@ pub struct BlockOutcome {
     pub events: Vec<Event>,
 }
 
+pub type ErrorCode = i64;
+pub enum CheckOutcome {
```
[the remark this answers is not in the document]
Byeongjee: You are right. There was no reason not to use `Result`.
I changed it.

core/src/client/importer.rs
```diff
@@ -240,8 +241,10 @@ impl Importer {
 
         // Enact Verified Block
         let db = client.state_db().read().clone(&parent.state_root());
+        let coordinator = Coordinator {};
 
-        let enact_result = enact(&block.header, &block.transactions, engine, client, db, &parent);
+        let enact_result =
+            enact(&block.header, &block.transactions, &block.evidences, engine, &coordinator, db, &parent);
```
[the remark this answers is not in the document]
Byeongjee: I merged it and separated "Remove block_fee interface" commit.

core/src/blockchain/blockchain.rs
```diff
@@ -401,18 +401,14 @@ pub trait BlockProvider: HeaderProvider + BodyProvider + InvoiceProvider {
     }
 
     /// Get transaction with given transaction hash.
-    fn transaction(&self, address: &TransactionAddress) -> Option<LocalizedTransaction> {
```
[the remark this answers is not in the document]
Byeongjee: I realized that this commit has some problem.
I wanted to replace `UnverifiedTransaction` and `VerifiedTransaction` types to `validator::Transaction` type in this commit, but it was not applied. So I fixed that.
I also realized that `LocalizedTransaction` does not have to be removed.
But `TransactionView` type should be removed because we don't read fields of Txes in the host side.

Byeongjee:
@sgkim126 Changes after your last review.
- I fixed all minor changes that you requested.
- I restored some methods and types that didn't have to be removed, but removed accidentally
    - possible_authors
    - LocalizedTransaction
    - PendingVerifiedTransaction

state/src/impls/test_helper.rs
sgkim126, line 19: These macros also should be moved into the module. I think it's good to remove them in these commits.

core/src/miner/mem_pool.rs
```diff
@@ -269,15 +269,103 @@ impl MemPool {
...
+
+        assert_eq!(mem_pool.transaction_pool.count, count);
+        assert_eq!(mem_pool.transaction_pool.mem_usage, mem_usage);
     }
 
     #[test]
     fn db_backup_and_recover() {
-        //TODO: Write test after implementing a mockup coordinato
-        todo!()
+        let validator = Arc::new(TestCoordinator::default());
```
sgkim126, line 341: Why don't you implement it in 92cb444287c161f2c4d4fa010236cc5ed107eeba?

coordinator/src/test_coordinator.rs
```diff
@@ -0,0 +1,96 @@
...
+        if transaction.size() > self.consensus_params.max_body_size() {
+            Err(1)
+        } else {
+            Ok(())
+        }
+    }
+
+    fn fetch_transactions_for_block<'a>(
+        &self,
+        transactions: Vec<&'a TransactionWithMetadata>,
```
sgkim126, line 66: Because the item type of the vector is a reference, it doesn't need to be a vector. How about using a slice?

coordinator/src/test_coordinator.rs
```diff
@@ -0,0 +1,96 @@
...
+            .map(|tx_with_metadata| TransactionWithGas {
+                tx_with_metadata,
+                gas: 0,
+            })
+            .collect()
+    }
+
+    fn remove_transactions<'a>(
+        &self,
+        transactions: Vec<&'a TransactionWithMetadata>,
```
sgkim126, line 79: ditto

core/src/miner/mem_pool.rs
sgkim126: I commented it because I think the test is not compiled in 92cb444287c161f2c4d4fa010236cc5ed107eeba. If it can be compiled, please ignore this comment.

Byeongjee:
> Why you don't include [deffc16](https://github.com/CodeChain-io/foundry/commit/deffc1639dd788d887d1ad9945bc0b61b84142d8) in here?

Are you asking why I didn't merge https://github.com/CodeChain-io/foundry/pull/260/commits/c3d85c861731acb3045a408a42cd6928c60b2360 and https://github.com/CodeChain-io/foundry/pull/260/commits/deffc1639dd788d887d1ad9945bc0b61b84142d8 into one?

That's because the miner unit tests required `TestCoordinator`.
I think the best practice was building the `TestCoordinator` before working on the miner code, but it was quite hard to reorder the commits. To reorder the commits, I had to reorder all validator-related commits, such as https://github.com/CodeChain-io/foundry/pull/260/commits/f3c7823509d850113bba2d1633feb41e800a545d

If you think it's intolerable to have such two commits, I'll try to reorder commits and merge them.

coordinator/src/test_coordinator.rs
Byeongjee: I used vector type to move references instead of copying them. 
The input references will be moved to inside of `TransactionWithGas` type.
I think it's not that meaningful. I'll use the slice.

core/src/miner/mem_pool.rs
Byeongjee: The answer is similar to the previous comment. Implementing `TestCoordinator` before this commit would be quite costly.

I didn't consider whether the test can be compiled or not, because even the main code was not compiling in the commit.
The commit would become too big if I had tried to make each commit compilable.

core/src/client/test_client.rs
```diff
@@ -196,25 +194,13 @@ impl TestBlockChainClient {
         header.set_parent_hash(*self.last_hash.read());
         header.set_number(n as BlockNumber);
         header.set_extra_data(self.extra_data.clone());
-        if let Some(addr) = author {
-            header.set_author(addr);
+        if let Some(author) = author {
```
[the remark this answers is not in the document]
Byeongjee: Fixed

state/src/impls/test_helper.rs
Byeongjee: Ok

state/src/item/validator_set.rs
```diff
@@ -0,0 +1,112 @@
...
+
+// TODO: implementation will be changed as we move NextValidators into a separate module
+pub struct NextValidatorSet(NextValidators);
+
+impl NextValidatorSet {
+    // Validator::close_block will return validator_set
+    // The host should update its next validators from the information
+    pub fn from_compact_validator_set(validators: CompactValidatorSet) -> Self {
+        let validators: Vec<_> = validators
+            .deref()
```
sgkim126, line 44: You don't need to call `deref`. The dot operator conducts an implicit dereferencing.

state/src/item/validator_set.rs
```diff
@@ -0,0 +1,112 @@
...
+    type Target = Vec<Validator>;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.deref()
+    }
+}
+
+impl From<NextValidatorSet> for Vec<SimpleValidator> {
+    fn from(validator_set: NextValidatorSet) -> Self {
+        validator_set.0.deref().iter().map(|val| SimpleValidator(*val)).collect()
```
sgkim126, line 74: ditto

state/src/item/validator_set.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// TODO: implementation will be changed as we move CurrentValidators into a separate module
+pub struct CurrentValidatorSet(CurrentValidators);
+
+impl CurrentValidatorSet {
+    pub fn create_compact_validator_set(&self) -> CompactValidatorSet {
+        CompactValidatorSet::new(
+            self.0
+                .deref()
```
sgkim126, line 85: ditto

core/src/miner/miner.rs
```diff
@@ -16,7 +16,7 @@
 
 use super::mem_pool::{Error as MemPoolError, MemPool};
 use super::{MinerService, MinerStatus};
-use crate::account_provider::Error as AccountProviderError;
```
sgkim126: A typo in the commit message.
minder -> miner

core/src/scheme/scheme.rs
```diff
@@ -137,6 +154,11 @@ impl Scheme {
         db.contains(&self.state_root())
     }
 
+    pub fn chain_initialized(&mut self, validators: CompactValidatorSet, consensus_params: ConsensusParams) {
```
sgkim126, line 157: I'm not sure about this change. IMO, if the genesis validators and consensus params are not described in the scheme file, `Scheme` should not have them.

sgkim126: I think you need to remove genesis_consensus_params and genesis_validators from Scheme.

coordinator/src/context.rs
```diff
@@ -1,3 +1,19 @@
+// Copyright 2020 Kodebox, Inc.
```
sgkim126, line 1: I think it's better to create a new PR for this commit.

coordinator/src/test_coordinator.rs
sgkim126: `rust
&'a [&'a TransactionWithMetadata]
`
Moving the reference and copying the reference is the same thing.

coordinator/src/context.rs
Byeongjee: Okay. How about the commits after this? Should they be moved to a separate PR as well?

core/src/scheme/scheme.rs
Byeongjee: I put `genesis_consensus_params` and `genesis_validators` in Scheme because I wanted to initialize state at once, by `Scheme::initialize_state`.
I agree that it's a bad idea to put the non-scheme fields to Scheme.

I'm looking for the best place to update `genesis_consensus_params` and `genesis_validators`.
Do you have any idea on this?

core/src/miner/miner.rs
Byeongjee: Thanks. I fixed it.

state/src/item/validator_set.rs
Byeongjee: Thanks. I removed it.

core/src/miner/miner.rs
Byeongjee: I fixed it.

core/src/event.rs
```diff
@@ -0,0 +1,61 @@
...
+    Block(BlockHash),
+    Transaction(TxHash),
+}
+
+#[derive(Clone, Default, PartialEq, Eq)]
+pub struct Events(pub Vec<Event>);
+
+impl Encodable for Events {
+    fn rlp_append(&self, _: &mut RlpStream) {
+        unimplemented!()
```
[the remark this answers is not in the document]
Byeongjee: Actually, no encoding scheme is decided yet, so it's okay to keep it unimplemented. 

However, I already implemented RLP encoding/decoding traits for other coordinator-related types in order to test some behavior, although the encoding scheme is subject to change.

Since I already implemented RLP for many other types, I think it's not a bad idea to implement it in Events type.
I'll add EventDB tests as well. (The reason I didn't implement RLP encoding for Events was that I didn't implement any EventDB tests)

coordinator/src/context.rs
Byeongjee: I'll move all commits after the license commit to a separate branch.

core/src/event.rs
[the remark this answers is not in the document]
Byeongjee: I added Encodable trait to Event and added some EventDB tests.

coordinator/src/context.rs
sgkim126: No. I meant only the license commit.

Byeongjee: No problem.

coordinator/src/test_coordinator.rs
Byeongjee: Okay. I'll fix it. Thanks.

core/src/client/importer.rs
```diff
@@ -263,10 +263,9 @@ impl Importer {
 
         // Enact Verified Block
         let db = client.state_db().read().clone(&parent.state_root());
-        let coordinator = &self.coordinator;
+        let validator = self.validator.borrow();
```
sgkim126:
```rust
let validator = &*self.validator;
```
It' the same code as yours. You don't need to use `Borrow` unless you want a generic code.

types/src/consensus_params.rs
```diff
@@ -17,6 +17,7 @@
 use cjson::scheme::Params;
 use ckey::NetworkId;
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+use std::str::FromStr;
```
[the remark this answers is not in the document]
Byeongjee: Thanks.

core/src/client/importer.rs
Byeongjee: Ok, thanks. I fixed both `miner` and `importer`

PR #262, opened 2020-03-13
Merge ClosedBlock and LockedBlock into one

### LINKED ISSUE #261: Remove ClosedBlock
There is no reason to distinguish `ClosedBlock` and `LockedBlock ` because `ClosedBlock` will never be reopened.
The reopen functionality was reserved for PoW engines, to support updating uncles.

Byeongjee:
Resolves #261 

There is no reason to distinguish ClosedBlock and LockedBlock since
ClosedBlocks are never reopened.
reopen methods should be removed as well.


sgkim126:
@majecty @HoOngEe  This PR will remove LockedBlock. After this commit, the ClosedBlock will not be reopened. I mentioned you just let you know.

PR #265, opened 2020-03-15
Fix security alert

sgkim126:
And this PR also makes the test project private to prevent mistakenly running the publish command.


test/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "author": "CodeChain Team <hi@codechain.io>",
   "license": "AGPL-3.0",
+  "private": true,
```
majecty: :+1:

PR #271, opened 2020-03-17
Convert hexadecimal string returned from getBalance to number

majecty:
Other getBalance callers are converting the return value to number using "+". Only this line does not have it.

### DIFF
--- a/test/src/e2e.long/staking.test.ts
+++ b/test/src/e2e.long/staking.test.ts
@@ -894,7 +894,7 @@ describe("Staking", function() {
                 address: author.toString(),
                 blockNumber: blockNumber - 1
             }))!;
-            const authorBalance = (await nodes[0].rpc.chain.getBalance({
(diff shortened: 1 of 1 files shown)


Byeongjee:
By the way, what happens if `getBalance` returns a string?
I hope this conversion would return a message informative enough to debug.

majecty:
@Byeongjee Currently, `getBalance` is returning string like "0x33". We are changing it to number using the "+" operator.

Byeongjee:
@majecty I was curious because `getBalance` can return `string`, `number`, or `null`.
It seems that it never returns a `number`. Maybe we should change this in foundry-rpc-js to reduce the ambiguity.

majecty:
Aha, you read the type. Since typescript's type cannot guaranty many things, you should check the code. Do not trust typescript's types. They can help you but they can fool you.

PR #275, opened 2020-03-18
Remove tracker from invoice and body db

Byeongjee:
This PR will remove tracker from transaction invoice and body db.
I didn't remove tracker entirely, because currently `ShardState` makes use of it.
Since shard will be removed, eventually tracker will be removed as well.
Remaining work should be done after `ShardState` is removed.

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -160,7 +160,6 @@ impl<'x> OpenBlock<'x> {
         }
 
         let hash = tx.hash();
-        let tracker = tx.tracker();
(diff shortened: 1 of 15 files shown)


sgkim126:
@Byeongjee Please resolve the conflict

Byeongjee:
@sgkim126 I force-pushed to resolve the conflict.

Byeongjee:
Another minor change.
I forgot to add test changes to the commit.

PR #277, opened 2020-03-18
Remove block rewards

sgkim126:
For now, the Foundry will burn the transaction fee. The fee distribution logic will be re-implemented in the module.


sgkim126:
@majecty Sorry for the inconvenience. I fixed one more test.

PR #280, opened 2020-03-19
Bump the minimist version

MSNTCS:


sgkim126:
If it's to fix the `minimist` security issue, it will not work, because `mkdirp` requests 0.0.8.
https://github.com/CodeChain-io/foundry/blob/d1841b7bf0b6a83cdbdb3357498ae3fce79b645f/test/yarn.lock#L1137

However, I think upgrading the library is good, so I don't disagree with this change.

PR #283, opened 2020-03-22
Remove HitHandler and replace ActionHandler to StakeHandler

sgkim126:
The hit handler is for testing custom actions. Since the Foundry will not support the custom actions, this handler is not necessary.

This PR also replaces the ActionHandler to the StakeHandler, but the handler id remains for SDK compatibility. Internally, the id is not used the transaction and RPC methods have the id fields.


sgkim126:
@majecty FYI, the only transaction that depends on core crate is ReportDoubleVote transaction, in other words, what StakeHandler does can be implemented in the state crate except for the transaction.
I'll replace StakeHandler to DoubleVoteHandler in https://github.com/CodeChain-io/foundry/pull/291.

PR #284, opened 2020-03-23
Make tests to check errors from foundry-rpc

majecty:


majecty:
@HoOngEe Could you check this commit: https://github.com/CodeChain-io/foundry/pull/284/commits/b5e117f6e56775a2691c2099838ad84b90ab1350 ?

There are two types of errors I couldn't resolve.
1. The test code is assuming that the network id is checked before the signature check, but currently, it is not.
2. The test code seems using incorrect length of signature.

HoOngEe:
Firstly, the second problem is due to the change in the `Rlp::Decodable` implementation. Previously, `ECDSA` signature decode was utilizing the `H520::decode` and it emits `RlpIsTooShort` or `RlpIsTooBig`. However, in my implementation I did not consider such a distinction so the implementation is like the [code](https://github.com/CodeChain-io/foundry/blob/408c02438f6f117bb6eceeff85dce58be94c5760/key/src/ed25519/signature.rs#L88). I did not check `Rlp::Decodable` for `H520` type so I did not consider such an error distinction.

HoOngEe:
For the first kind of the problem, I'll investigate it.

HoOngEe:
For the first kind of the problem, it is inevitable to modify the test code.
When a transaction is sent through `SendSignedTransaction`, signature verification is called. Previously, that is `recover_public` and now it is `verify`. However, `recover_public` can succeed even if the signature is not generated form the target message. Therefore, CodeChain can get into the network id verification step. However, Foundry fails in signature verification because the message was modified.
The following proves the concept of my scenario:
```
// The code from "verification.test.ts"
let encoded: any[];
        beforeEach(async function() {
            const seq = await node.sdk.rpc.chain.getSeq(faucetAddress);
            const tx = node.sdk.core
                .createPayTransaction({
                    recipient,
                    quantity: 0
                }).sign({
                    secret: faucetSecret,
                    fee: 10,
                    seq
                });
            encoded = tx.toEncodeObject();
            let prev_msg = tx.unsigned.unsignedHash().toString();
            console.log(prev_msg);
            const encodedd = tx.unsigned.toEncodeObject();
            encodedd[2] = "ac";
            let new_msg = blake256(RLP.encode(encodedd));
            console.log(new_msg);
            console.log(encoded[4]);
            console.log(recoverEcdsa(new_msg, encoded[4]));
            console.log(recoverEcdsa(prev_msg, encoded[4]));
        });
```
```
standard output:
ecb2d86987e1d24e66553e1f8f90a5407466a3f5165bd688fc095d78e6a0f153 // original message hash
768f6690660fd605ff9a19f894780b9fc7514d9930aa11c5341bcbe84045112e // modified message hash
0x27041556d6057ce167dcec534fff8a50be89fa764d09d4c29ee6f1ac9545fdf078be0d5bb4754ef80dad2eeed8e0296d7b3656a24a94c4a01517e81549b2850500 // Common signature
e9476055310154098d7e852d8ade2594eb255db3dc79b825f67c86c9d23dc4a41714c8c348e9ebeb06ddc8a34f753944fc0a737b5f016c1c2d92eff3a8240232 // recovered public key from the modified message 
55f2c44106dc980941313e0a0a8b7313ca70da62484fdae17087ef3ee318a72d155541a183824a002b02b3855e9d223f18ba803dc62370a07852ba6df4a8b9a6 // recovered public key from the original message which is same with faucetPublic
```
Therefore, we should modify the test code.

majecty:
@HoOngEe Thanks to your help, I fixed verification tests.

PR #285, opened 2020-03-23
Cold events for Informer API

MSNTCS:


majecty:
Please do not update the `main` function. Since we are using the async code only in the Informer code, please do not expose the async code outside of the Informer. You can use `tokio::runtime`.
- https://docs.rs/tokio/0.2.13/tokio/runtime/index.html 

For the hot service case, receiving messages from the message queue is a blocking call. IMO, we don't need to use the async function for the hot events. In the WIP PR,  the hot_service function's type is async, but the behavior is sync. It makes confusing.

In the cold events, how about making each cold event generator have a loop? Then we don't need to use lock to generate messages.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +121,68 @@ impl InformerService {
...
+            client,
+            rate_limiter: ratelimit::Builder::new().quantum(1).interval(Duration::new(0, 5000)).build(),
+
+            send_status: ColdEventSendStatus::InProgress,
+        }
+    }
+
+    pub fn run(mut self, connection: Connection, mut favorite_block_number: u64) {
+        println!("I am here 3");
+        let rt = Runtime::new().unwrap();
```
majecty, line 143: You should not use multiple runtimes if you don't have any special purpose on it.
Please use only one runtime for the Informer service.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +121,68 @@ impl InformerService {
...
+        rt.spawn(async move {
+            task::spawn(async move {
+                loop {
+                    let event = self.gen(favorite_block_number);
+                    connection.cold_notification(&event);
+                    self.rate_limiter.wait();
```
majecty, line 149: You should not call `blocking` function in an async function.
The `rate_limiter.wait` is a blocking call. You must not use it.
You can implement a simple rate limiter on your own. Or you can search for a rate limiter that can be used with async functions.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +121,68 @@ impl InformerService {
...
+            let current_id = BlockId::Number(favorite_block_number);
+            let block = self.client.block(&current_id).map(|block| {
+                let block = block.decode();
+                RPCBlock::from_core(block, self.client.network_id())
+            });
+            let current_block = block
+                .expect("In the process of generating block generation cold event response, the block is not created");
+            event = ColdEvents::BlockGeneration(Box::new(current_block));
+            if favorite_block_number == best_block_number {
+                self.send_status = ColdEventSendStatus::AllEventsSent;
```
majecty, line 183: Changing self.send_status in the `gen` function is awkward. When you write a function, please consider whether this function can modify variables. When you make a function to modify the self's state, you should give a meaningful name or concept to the function.

When I read the function's name `gen` i don't expect the function to modify the self's state. Please do not modify any state in this function. Please read self's state, and return generated events.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +121,68 @@ impl InformerService {
...
+        println!("I am here 3");
+        let rt = Runtime::new().unwrap();
+        rt.spawn(async move {
+            task::spawn(async move {
+                loop {
+                    let event = self.gen(favorite_block_number);
+                    connection.cold_notification(&event);
+                    self.rate_limiter.wait();
+                    match self.send_status {
+                        ColdEventSendStatus::AllEventsSent => {
```
majecty, line 151: Should we need `AllEventSent` for the "BlockGenerated" event?
Block will be generated forever. We ca nsend blocks forever to the client.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -67,7 +81,15 @@ impl InformerService {
                     index if index == connection_index => {
                         if let Ok(connection) = connection_rcv.recv() {
                             cinfo!(INFORMER, "A new connection is added");
-                            self.add_new_connection(connection);
+                            self.add_new_connection(connection.clone());
+                            if let ColdEventStatus::Enable = &connection.has_cold_event {
```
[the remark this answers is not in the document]
MSNTCS: If the connection does not have a cold event subscription we just add the connection. Otherwise, prepare the response of the cold event.

informer/src/informer_service/informer_service_handler.rs
MSNTCS: I do not use multiple run time. it is the only place that I used `runtime`.

informer/src/informer_service/informer_service_handler.rs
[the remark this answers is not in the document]
majecty: The same effect can be done from the `if let EventTags::ColdBlockGenerationNumerical(value) = interested_events` code below.
IMO, removing `if let ColdEventStatus::Enable = &connection.has_cold_event {` will not change the current behavior.

informer/src/informer_service/informer_service_handler.rs
MSNTCS: `AllEventsSent` is necessary for the time that there is no new block un our case. If we want to inform the client continuously, I need to think of a procedure.

majecty: We should give new block information forever.
For example, the best block is 5 and a client wants to receive new blocks from block number 3.
Foundry should give blocks 3, 4, 5 from disk. Foundry should give blocks 6, 7, 8, ... whenever blocks are generated.

informer/src/informer_service/informer_service_handler.rs
majecty: The position is problem.
If we want to add `TransactionMinedEventGenerator`, where will be the runtime?
If we follow the current structure, there will be multiple runtimes; one runtime for one generator.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -102,3 +116,46 @@ impl InformerService {
...
+
+    pub fn run(mut self, connection: Connection, mut favorite_block_number: u64) -> tokio::task::JoinHandle<()> {
+        task::spawn(async move {
+            loop {
+                let best_block_number = self.client.best_block_header().number();
+                if best_block_number >= favorite_block_number {
+                    let event = self.gen(favorite_block_number);
+                    connection.cold_notification(&event);
+                    self.async_runtime_wait().await;
+                    favorite_block_number += 1;
```
majecty, line 145: Please call sleep when the best_bock_number < favorite_block_number.

```suggestion
                let best_block_number = self.client.best_block_header().number();
                if best_block_number >= favorite_block_number {
                    let event = self.gen(favorite_block_number);
                    connection.cold_notification(&event);
                    self.async_runtime_wait().await;
                    favorite_block_number += 1;
                } else {
                    sleep();
                }   
```

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -102,3 +116,46 @@ impl InformerService {
...
+    }
+
+    fn gen(&mut self, favorite_block_number: u64) -> ColdEvents {
+        let current_id = BlockId::Number(favorite_block_number);
+        let block = self.client.block(&current_id).map(|block| {
+            let block = block.decode();
+            RPCBlock::from_core(block, self.client.network_id())
+        });
+        let current_block =
+            block.expect("In the process of generating block generation cold event response, the block is not created");
```
majecty, line 158: The reason `block != None` is that you check `best_block_number >= favorite_block_number` before calling `gen` function.

```suggestion
            block.expect("We checked favorite_block_number <= best_block_number before calling the function");
```

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -67,7 +75,16 @@ impl InformerService {
                     index if index == connection_index => {
                         if let Ok(connection) = connection_rcv.recv() {
                             cinfo!(INFORMER, "A new connection is added");
-                            self.add_new_connection(connection);
+                            self.add_new_connection(connection.clone());
+                            let client = self.client.clone();
+                            rt.spawn(async move {
+                                for interested_events in connection.interested_events.clone() {
```
majecty, line 81: Do we nee clone here?

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -67,7 +75,16 @@ impl InformerService {
...
+                                    if let EventTags::ColdBlockGenerationNumerical(value) = interested_events {
+                                        let cold_generator = BlockCreatedEventGenerator::new(client.clone());
```
majecty, line 83: Do we need to clone client here?

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -102,3 +116,46 @@ impl InformerService {
...
+            client,
+            rate_limiter: ratelimit::Builder::new().quantum(1).interval(Duration::new(0, 5000)).build(),
+        }
+    }
+
+    pub async fn async_runtime_wait(&mut self) {
+        self.rate_limiter.wait();
+    }
+
+    pub fn run(mut self, connection: Connection, mut favorite_block_number: u64) -> tokio::task::JoinHandle<()> {
```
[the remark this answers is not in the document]
majecty: Could you change the name of `favorite_block_number`? How about `fromBlockNumber`?

informer_courier/src/informer_notify.rs
```diff
@@ -13,6 +13,7 @@
 //
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
```
majecty, line 16: Please remove this empty line.

informer/src/informer_service/informer_service_handler.rs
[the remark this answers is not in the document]
MSNTCS: Yes, I need `favorit_block_number` to because the function changes its value.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -102,3 +116,46 @@ impl InformerService {
...
+impl BlockCreatedEventGenerator {
+    pub fn new(client: Arc<Client>) -> Self {
+        Self {
+            client,
+            rate_limiter: ratelimit::Builder::new().quantum(1).interval(Duration::new(0, 5000)).build(),
+        }
+    }
+
+    pub async fn async_runtime_wait(&mut self) {
+        self.rate_limiter.wait();
```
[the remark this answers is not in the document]
MSNTCS: If I used it in async function would it be still a blocker?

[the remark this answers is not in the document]
majecty: Yes blocking function always blocks. It doesn't matter where you called it.

informer/src/informer_service/informer_service_handler.rs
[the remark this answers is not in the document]
majecty: If you want to modify a variable inside a function, please declare it in the function.
If you make the function argument as mutable, you are modifying the variable of the outside of the function.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +116,49 @@ impl InformerService {
...
+        })
+    }
+
+    fn gen(&mut self, favorite_block_number: u64) -> ColdEvents {
+        let current_id = BlockId::Number(favorite_block_number);
+        let block = self.client.block(&current_id).map(|block| {
+            let block = block.decode();
+            RPCBlock::from_core(block, self.client.network_id())
+        });
+        let current_block = block.unwrap();
```
majecty, line 161: Please use `expect` instead of `unwrap`

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +116,49 @@ impl InformerService {
...
+impl BlockCreatedEventGenerator {
+    pub fn new(client: Arc<Client>) -> Self {
+        Self {
+            client,
+            rate_limiter: LeakyBucket::builder()
+                .max(100)
+                .refill_interval(Duration::from_secs(1))
+                .refill_amount(100)
+                .build()
+                .unwrap(),
```
majecty, line 134: Is `unwrap` safe? If you think it is safe, please describe it using `expect`.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -14,33 +14,39 @@
...
-    PeerAdded(String, String, usize),
+pub enum ColdEvents {
+    None,
+    BlockGeneration(Box<RPCBlock>),
+    SendingDone(String),
 }
 
 pub struct InformerService {
-    connections: Vec<Connection>,
+    connections: Vec<Arc<Connection>>,
```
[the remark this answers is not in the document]
MSNTCS: I think because I am using the connection in different threads, Arc help to avoid cloning it.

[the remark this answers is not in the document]
majecty: Why I said "Remove clone" is that there was no need to use "clone". In most cases, you can use a reference instead of calling clone.
Using `Arc` to avoid cloning is awkward.

You should consider whether you will use Arc or not when you design the `Connection` type.

What I feel awkward is this point. If you want to make only one Connection instance for real connection, you can make `Connection` not Clonable and only use it with `Arc`.

If you want to make the ColdEventGenerator and the InformerService have its own Connection object for the same real connection, you should not use `Arc` and make the `Connection` type Clonable.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -98,3 +116,49 @@ impl InformerService {
...
+
+    pub fn run(mut self, connection: Arc<Connection>, from_block_number: u64) -> tokio::task::JoinHandle<()> {
+        let mut favorite_block_number = from_block_number;
+        task::spawn(async move {
+            loop {
+                let best_block_number = self.client.best_block_header().number();
+                if best_block_number >= favorite_block_number {
+                    let event = self.gen(favorite_block_number);
+                    connection.cold_notify(&event);
+                    self.rate_limiter.acquire_one().await.expect("Responses are more than the limit");
```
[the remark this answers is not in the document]
MSNTCS: I think the error happens when there is no token to acquire. This happens when we reach the limit.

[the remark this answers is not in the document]
majecty: If there is no token to acquire, the rate_limiter should not throw an error. It should wait until a token will be generated. If you think the library throws an error when there is no token, we should not use the library.

informer/src/handler/connection.rs
```diff
@@ -63,10 +63,27 @@ impl Connection {
...
+                }
+            },
+        }
+    }
+
     pub fn notify_client(&self, event: &Events) {
         let json_object = serde_json::to_value(event).expect("json format is not valid").as_object_mut().cloned();
         let params = Params::Map(json_object.expect("Event is serialized as object"));
         match self.status {
+            // FIXME : We should not use wait and .await. The standard Future is not supported by the current paritytech/jsonrpc crate.
```
majecty, line 85: ditto.

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -99,3 +114,52 @@ impl InformerService {
...
+        task::spawn(async move {
+            loop {
+                let best_block_number = self.client.best_block_header().number();
+                if best_block_number >= current_block_number {
+                    let event = self.gen(current_block_number);
+                    connection.cold_notify(&event);
+                    self.rate_limiter
+                        .acquire_one()
+                        .await
+                        .expect("Note that you are not allowed to acquire more tokens than the current");
```
majecty, line 147: What is the meaning of "Note that you are not allowed to acquire more tokens than the current"?

informer/src/informer_service/informer_service_handler.rs
```diff
@@ -99,3 +114,52 @@ impl InformerService {
...
+        })
+    }
+
+    fn gen(&mut self, favorite_block_number: u64) -> ColdEvents {
+        let current_id = BlockId::Number(favorite_block_number);
+        let block = self.client.block(&current_id).map(|block| {
+            let block = block.decode();
+            RPCBlock::from_core(block, self.client.network_id())
+        });
+        let current_block = block.expect("There is no block with the calculated Block ID");
```
majecty, line 162: "There is no block with the calculated Block ID" is not  the reason that you can call `expect` safely.
You should give the reason why you can call `expect` safely.
You wrote an "error message". It is not a proof or reason.

informer/src/informer_service/rate_limiter.rs
```diff
@@ -0,0 +1,40 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::time::Duration;
+use tokio::time;
+
+pub struct RateLimiter {
+    ticket: u64,
+    pub current_ticket: u64,
```
majecty, line 22: Could you tell me why the current_ticket should be public?

informer/src/informer_service/rate_limiter.rs
```diff
@@ -0,0 +1,40 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::time::Duration;
+use tokio::time;
+
+pub struct RateLimiter {
+    ticket: u64,
```
majecty, line 21: How about `maximum_ticket`?

informer/src/informer_service/rate_limiter.rs
```diff
@@ -0,0 +1,40 @@
...
+            ticket,
+            current_ticket: 0,
+        }
+    }
+
+    pub async fn acquire_ticket(&mut self) {
+        self.current_ticket += 1;
+        if self.current_ticket == self.ticket {
+            self.current_ticket = 0;
+            time::delay_for(Duration::from_millis(1000)).await;
```
majecty, line 37: Current implementation sleeps 1 second every N tickets.
How about sleeping if tickets acquired more than maximum in a second?

majecty: For example

```rs
self.current_ticket += 1;
let remain
let elapsed = self.last_reset.elapsed();
if elapsed < Duration::from_secs(1) && self.current_ticket == self.maximum_ticket {
  time::delay_for(Duration::from_secs(1) - elapsed).await;
}
```

informer/src/informer_service/rate_limiter.rs
```diff
@@ -0,0 +1,46 @@
...
+            last_reset: Instant::now(),
+        }
+    }
+
+    pub async fn acquire_ticket(&mut self) {
+        self.current_ticket += 1;
+        let elapsed = self.last_reset.elapsed();
+        if elapsed < Duration::from_secs(1) && self.current_ticket == self.maximum_ticket {
+            time::delay_for(Duration::from_secs(1) - elapsed).await;
+            self.current_ticket = 0;
```
majecty, line 40: self.last_reset = Instant::now();

PR #286, opened 2020-03-24
Rename ChainUpdateResult variables

Byeongjee:
This should have been done in https://github.com/CodeChain-io/foundry/pull/256


core/src/client/importer.rs
```diff
@@ -399,10 +399,11 @@ impl Importer {
         let chain = client.block_chain();
 
         let mut batch = DBTransaction::new();
-        let route = chain.insert_header(&mut batch, &HeaderView::new(&header.rlp_bytes()), self.engine.borrow());
+        let update_results =
```
majecty: Shouldn't we use `update_result` here?

PR #288, opened 2020-03-26
Implement Account module

somniumism:
I wrote a [design](https://www.notion.so/codechain/Module-Account-Interface-40877ac516b941b4982f86fde4301b87) for module Account. Depending on the model, I will implement an interface with four methods that other modules can access.

I removed the logic related to `Account` from the foundry.

#### Work that remove unnecessary logic related to account from the foundry
- [x] Remove the trait `AccountData` from **client** in the foundry
- [x] Remove the struct `AccountDetails` from **miner** in the foundry
- [x] Remove the struct `Account` from **state** in the foundry
        - Remove logic related to the **account** from the top level state
- [x] Remove the struct `PodAccount` from **scheme** in the foundry
        - Remove `pod_account.rs` and `pod_state.rs`
        - Remove `Account` from **json**
- [x] Turn off the e2e tests related to this work

#### The remaining logic
- `AccountProvider` in **consensus**
- `StakeAccount` in **consensus**
- `StakeAccount` in **json**
- Logic related to the account in **rpc**
- `StakeAccount` in **state**


basic_module/account/src/account/checktx_handler.rs
```diff
@@ -0,0 +1,48 @@
+// use super::view::{get_balance, get_sequence};
+use super::internal::get_storage_id;
+use crate::{get_context, Account, Action, Address, Transaction};
+
+pub trait CheckTxHandler {
```
HoOngEe, line 5: To comply with FML rules, we should make one explicit trait implementation rather than default implementation.

basic_module/account/src/account/executer.rs
```diff
@@ -0,0 +1,72 @@
+use super::error::Error;
+use super::internal::get_storage_id;
+use crate::{get_context, Account, Action, Address, Transaction};
+use coordinator::context::SubStorageAccess;
+
+pub trait TransactionExecuter {
+    fn execute_transactions(context: &mut dyn SubStorageAccess, transactions: &[Transaction]) {
```
HoOngEe, line 7: ditto.

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,220 @@
...
+        match &tx.action {
+            Action::Pay {
+                sender,
+                receiver: _,
+                quantity,
+            } => {
+                let sender_balance = get_balance(&sender);
+                let sequence = get_sequence(&sender);
+                if sender_balance.is_none()
+                    && sequence.is_none()
```
HoOngEe, line 23: The clauses should be linked with logical OR

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,220 @@
...
+                    && sender_balance.unwrap() < tx.fee + quantity
+                    && sequence.unwrap() <= tx.seq
+                {
+                    return false
```
HoOngEe, line 27: If what you want to test is the boolean expression and the determined value is also boolean, It seems clear to return the test formula directly

HoOngEe:
I think we need `increase_seq` in the `AccountReviser` trait.

state/src/impls/top_level.rs
```diff
@@ -383,12 +374,9 @@ impl TopLevelState {
     ) -> StateResult<()> {
         match action {
             Action::Pay {
-                receiver,
-                quantity,
-            } => {
-                self.transfer_balance(sender_address, receiver, *quantity)?;
-                Ok(())
-            }
+                receiver: _,
```
HoOngEe, line 377: It is just a comment but we were using the pattern `Action::Pay {..}` when ignoring some fields of structs.

state/src/impls/top_level.rs
```diff
@@ -618,533 +553,6 @@ impl TopState for TopLevelState {
         Ok(())
     }
 }
-
-#[cfg(test)]
```
HoOngEe, line 622: I think we should not naively remove some tests. Though the tests are related to state changes due to Account-related transactions, `clone`, `override_cache`, etc are closely related to general db operations.

HoOngEe: If some suitable transactions are not available from the top_level, let's apply module level state changes and examine the same tests onto it.

state/src/stake/actions.rs
```diff
@@ -65,7 +65,7 @@ pub fn init_stake(
     stakeholders.save_to_state(state)?;
 
     for (address, deposit) in &genesis_candidates {
-        let balance = state.balance(address).unwrap_or_default();
+        let balance: u64 = Default::default();
```
HoOngEe, line 68: How about leaving some comments in the code to give a piece of information about the counterintuitive code?

state/src/stake/actions.rs
```diff
@@ -975,15 +968,15 @@ mod tests {
...
         init_stake(&mut state, Default::default(), Default::default(), Default::default()).unwrap();
 
         // TODO: change with stake::execute()
         let result = self_nominate(&mut state, &address, &address_pubkey, 0, 0, 5, b"metadata1".to_vec());
         assert_eq!(result, Ok(()));
-
-        assert_eq!(state.balance(&address).unwrap(), 1000);
+        
+        let balance: u64 = 1000;
+        assert_eq!(balance, 1000);
```
HoOngEe, line 979: I think these assertions are unnecessary.

core/src/scheme/mod.rs
```diff
@@ -15,8 +15,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 mod genesis;
-mod pod_account;
```
HoOngEe, line 18: If you remove the `pod_account` then you can remove the serialized initial accounts information from `.json` scheme files.

HoOngEe:
I think you need to cleanup the commits not to erase some code after creating the code in a single pr.

Byeongjee:
@somniumism Did you decide to put banking functionality in the Account module?

somniumism:
Yes, I did. Functionally, this module contains the bank module of the Cosmos SDK.

somniumism:
@HoOngEe  When the review and its modifications are completed, I will clean up many commits of this PR.

Byeongjee:
@somniumism What do you mean by turning of e2e tests related to this work?
I think none of the tests will work until we connect the account module to the host with the coordinator.

test/src/e2e.long/staking.test.ts
```diff
@@ -388,58 +388,61 @@ describe("Staking", function() {
         );
     }).timeout(60_000);
 
-    it.skip("doesn't leave zero balance account after transfer", async function() {
```
HoOngEe: What is this commit with the name . ?

somniumism: It's my mistake, I fixed it.

basic_module/account/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    pub fee: u64,
+
+    pub network_id: NetworkId,
+
+    pub action: Action,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Hash)]
+pub struct NetworkId([u8; 2]);
```
HoOngEe, line 64: I think we should reconsider about this `NetworkId` type data. Of course module can define this kind of data but in this module I could not find the actual usage of this data. Previously, `PlatformAddress` is constructed using `Address` and `NetworkId` but removed here.

somniumism: Thank you for your good advice. I modified it.

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,150 @@
...
+                quantity,
+            } => {
+                let sender_balance = get_balance(&sender);
+                let sender_sequence = get_sequence(&sender);
+                match (sender_balance, sender_sequence) {
+                    (Some(balance), Some(sequence)) => !(balance < tx.fee + quantity || sequence <= tx.seq),
+                    _ => false,
+                }
+            }
+            _ => false,
```
HoOngEe, line 26: panic!() seems okay because wrongly routed transaction means a failure in our coordinator.

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,150 @@
...
+            match &tx.action {
+                Action::Pay {
+                    sender,
+                    receiver,
+                    quantity,
+                } => {
+                    let sender_balance = get_previous_balance(context, &sender).unwrap();
+                    let sender_sequence = get_previous_sequence(context, &sender).unwrap();
+                    if sender_balance >= tx.fee + quantity || sender_sequence > tx.seq {
+                        let _ = sub_balance(&sender, tx.fee + quantity);
```
HoOngEe, line 46: `sub_balance` checks if `sender_balance >= tx.fee + quantity` . Instead of double checking, let's utilize the result type.

basic_module/account/src/account/executer.rs
```diff
@@ -1,57 +1,31 @@
...
+        return Err(AccountError::InvalidValue)
     }
 
     account.balance -= val;
     context.set(get_storage_id(), address, account.to_vec());
     Ok(())
 }
 
-fn get_previous_balance(context: &mut dyn SubStorageAccess, address: &Address) -> Option<u64> {
+pub fn get_previous_balance(context: &mut dyn SubStorageAccess, address: &Address) -> Option<u64> {
```
HoOngEe, line 28: It just fetches current balance however the name is `get_previous_balance`

basic_module/account/src/traits/core.rs
```diff
@@ -0,0 +1,38 @@
...
+
+        fn increase_sequence(address: &Address) -> Result<(), AccountError>;
+    }
+
+    pub trait AccountView {
+        fn has_account(address: &Address) -> bool;
+
+        fn is_active(address: &Address) -> Result<bool, AccountError>;
+
+        fn get_balance(context: &mut dyn SubStorageAccess, address: &Address) -> Option<u64>;
```
HoOngEe, line 30: The interface should not expect the context as an input, account data is managed by `Account` module so that must be resolved internally.

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,150 @@
...
+            .get(get_storage_id(), address)
+            .map(|_| Account {
+                sequence: 0,
+                balance: 0,
+            })
+            .map(|account| account.balance)
+    }
+
+    fn get_sequence(context: &mut dyn SubStorageAccess, address: &Address) -> Option<u64> {
+        context
```
HoOngEe, line 133: Why don't you reuse the module scope function already implemented in the `executer`?

core/src/consensus/stake/mod.rs
```diff
@@ -193,7 +193,6 @@ mod tests {
 
         let mut state = metadata_for_election();
         increase_term_id_until(&mut state, 29);
-        state.add_balance(&address, 1000).unwrap();
```
HoOngEe, line 196: I think ignoring the test is better than executing the meaningless code. You can ignore the test function using the attribute "ignore". Let's remove these test when implementing stake modules.
```
#[test]
#[ignore]
```

core/src/miner/mem_pool.rs
```diff
@@ -1084,20 +1084,22 @@ pub mod test {
         MemPoolInput::new(signed, TxOrigin::Local)
     }
 
-    #[test]
+    #[allow(dead_code)]
```
HoOngEe, line 1092: ditto. #[ignore]

rpc/src/v1/impls/chain.rs
```diff
@@ -65,15 +65,21 @@ where
     }
 
     fn get_seq(&self, address: PlatformAddress, block_number: Option<u64>) -> Result<Option<u64>> {
-        let block_id = block_number.map(BlockId::Number).unwrap_or(BlockId::Latest);
-        let address = address.try_address().map_err(errors::core)?;
-        Ok(self.client.seq(address, block_id))
+        let _block_id = block_number.map(BlockId::Number).unwrap_or(BlockId::Latest);
```
HoOngEe, line 71: Is there any problem with removing these functions?

basic_module/account/src/key/src/network.rs
```diff
@@ -0,0 +1,106 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+use serde::de::Error;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
+use std::str::{self, FromStr};
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Hash)]
+pub struct NetworkId([u8; 2]);
```
Byeongjee, line 24: @somniumism Why did you define `NetworkId` on your own?
Since `NetworkId` is a part of `ConsensusParams`, which is managed by the host, I think you shouldn't define this by yourself.

Byeongjee: Why do you have a separate `key` module in your crate?

basic_module/account/src/account/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod checktx_handler;
+mod error;
+pub mod executer;
```
Byeongjee, line 3: executor

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn set_balance(address: &Address, val: u64) -> Result<(), AccountError> {
+        let context = get_context();
+        let mut account: Account = context.get(get_storage_id(), address).ok_or(AccountError::InvalidAddress)?.into();
+
+        account.balance = val;
+        context.set(get_storage_id(), address, account.to_vec());
+        Ok(())
+    }
+
+    fn increase_sequence(address: &Address) -> Result<(), AccountError> {
```
Byeongjee, line 93: IMO `incremet_sequence` is better

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,150 @@
...
+                Action::Custom {
+                    ..
+                } => {}
+            }
+        }
+    }
+}
+
+#[allow(dead_code)]
+pub struct Reviser {}
```
Byeongjee, line 59: Do you have any better name for this?
@somniumism @HoOngEe

Byeongjee: @somniumism Why not just use `AccountManager`?

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,117 @@
...
+impl TransactionExecutor for Executor {
+    fn execute_transactions(transactions: &[Transaction]) {
+        for tx in transactions {
+            match &tx.action {
+                Action::Pay {
+                    sender,
+                    receiver,
+                    quantity,
+                } => {
+                    let _ = sub_balance(&sender, tx.fee + quantity).and_then(|_| add_balance(&receiver, *quantity));
```
HoOngEe, line 41: Let's add a comment above this line like "FIXME: Suitable error handling is needed". I'm not sure about storing errors, so let's just leave a comment for a later modification.

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+}
+
+#[allow(dead_code)]
+pub struct Manager {}
+
+impl AccountManager for Manager {
+    fn add_balance(address: &Address, val: u64) -> Result<(), AccountError> {
+        add_balance(address, val)
```
HoOngEe, line 56: Someone can misunderstand this as a recursive call so that let's handle this issue with `use ${path_to_add_balance}::add_balance as core_add_balance`

basic_module/account/src/traits/core.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    pub trait AccountView {
+        fn is_active(address: &Address) -> bool;
+
+        fn get_balance(address: &Address) -> u64;
+
+        fn get_sequence(address: &Address) -> u64;
+    }
+
+    pub trait SignatureManger {
```
HoOngEe, line 33: Typo here, `Manger` -> `Manager`

basic_module/account/src/traits/core.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub trait CheckTxHandler {
+        fn verify_transaction(tx: &Transaction) -> bool;
+    }
+
+    pub trait TransactionExecutor {
+        fn execute_transactions(transactions: &[Transaction]);
+    }
+
+    pub trait AccountManager {
+        fn add_balance(address: &Address, val: u64);
```
HoOngEe, line 16: Foundry module system requires interfaces to have (&self) parameter because it utilizes `trait Object`.

basic_module/account/src/traits/core.rs
somniumism: Thanks, I fixed it.

basic_module/account/src/traits/core.rs
somniumism: I've fixed it on previous modification, I didn't update it. Thanks. I just updated it.

HoOngEe:
@somniumism I think we need an additional method `sign` which creates signature bytes from `message bytes` and `secret key`, could you add this function?

basic_module/account/src/traits/impls.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[allow(dead_code)]
+pub struct SigManager {}
+
+impl SignatureManager for SigManager {
+    fn verify(&self, signature: &Signature, message: &[u8], public: &Public) -> bool {
+        verify_ed25519(signature, message, public)
+    }
+
+    fn sign(&self, message: &[u8], private: &Private) -> Signature {
```
somniumism, line 120: @HoOngEe I added a method.

basic_module/account/src/account/internal.rs
```diff
@@ -0,0 +1,52 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use super::AccountError;
+use crate::{get_context, get_storage_id, Account};
+use key::Address;
+
+pub fn add_balance(address: &Address, val: u64) {
+    let context = get_context();
```
HoOngEe, line 22: How about adding `val == 0` check to reduce storage access when any change is not expected like the previous `TopLevelState` operation.

HoOngEe:
Additional Todos:
- [ ] Accumulate fee to intermediate rewards when executing transactions
- [ ] Return block outcome when `end_block` based on the accumulated transaction results

basic_module/account/src/account/internal.rs
somniumism: I added!

basic_module/account/src/types.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[allow(dead_code)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub enum Action {
+    Pay {
+        sender: Public,
+        receiver: Public,
+        quantity: u64,
+    },
+    Custom {
```
HoOngEe, line 82: Let's remove the `Custom` action which is of no use.

basic_module/account/src/import.rs
```diff
@@ -0,0 +1,19 @@
...
+pub trait FeeManager {
+    fn accumulate_block_fee(&self, total_additional_fee: u64, total_min_fee: u64);
+}
+
+#[allow(dead_code)]
+pub fn fee_manager() -> Box<dyn FeeManager> {
+    unimplemented!()
+}
+
+pub trait SignatureManager {
```
HoOngEe, line 12: `SignatureManager` is not imported from other modules but provided by this module. BTW in the current FML, we do not need such a distinction: exported or imported.

basic_module/account/src/impls.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl CheckTxHandler for Handler {
+    fn verify_transaction(&self, tx: &Transaction) -> bool {
+        match &tx.action {
+            Action::Pay {
+                sender,
+                receiver: _,
+                quantity,
+            } => !(get_balance_internally(&sender) < tx.fee + quantity || get_sequence_internally(&sender) <= tx.seq),
+            _ => panic!("This is a wrong transaction format that the module account couldn't handle."),
```
HoOngEe, line 38: After removing the `Custom` type, the wildcard pattern matching can be removed.

basic_module/account/src/impls.rs
```diff
@@ -39,3 +39,53 @@ impl CheckTxHandler for Handler {
...
+
+#[allow(dead_code)]
+pub struct Executor {}
+
+impl TransactionExecutor for Executor {
+    fn execute_transactions(
+        &self,
+        _context: &mut dyn SubStorageAccess,
+        transactions: &[SignedTransaction],
+    ) -> Result<Vec<TransactionExecutionOutcome>, ErroneousTransactions> {
```
HoOngEe, line 51: I modified the return type of `execute_transactions` to return Result<Vec<TransactionExecutionOutcome>, ()>. Let's implement the module fit into this type before a suitable host-side handling for ErroneousTransactions is equipped.

basic_module/account/src/core.rs
```diff
@@ -14,11 +14,19 @@
...
 pub use coordinator::types::TransactionExecutionOutcome;
 
 pub trait CheckTxHandler {
     fn verify_transaction(&self, tx: &Transaction) -> bool;
 }
+
+pub trait TransactionExecutor {
+    fn execute_transactions(
+        &self,
+        context: &mut dyn SubStorageAccess,
```
HoOngEe, line 30: Let's remove the argument context here in the module side.

basic_module/account/src/types.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct Transaction {
+    pub seq: u64,
+    pub fee: u64,
+    pub network_id: NetworkId,
+    pub action: Action,
+}
+
+impl Transaction {
+    pub fn hash(&self) -> H256 {
```
HoOngEe, line 61: Is there an actual usage for this method in the module?

somniumism: As I know, It's used for verifying signature when transaction is executed.

basic_module/account/src/impls.rs
somniumism: It means that we don't need `ErroneousTransactions`? If it returns `Result<Vec, ()>`, what should the module return?

basic_module/account/src/import.rs
somniumism: Oh.. This is a huge mistake. Thank you.

basic_module/account/src/types.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Clone)]
+pub struct Account {
+    pub balance: u64,
+    pub sequence: u64,
+}
+
+// Temporary measures for build.
+// This function will be removed when the structure of the context and conversion logic is completed.
+impl From<Vec<u8>> for Account {
+    fn from(_vec: Vec<u8>) -> Account {
```
HoOngEe, line 17: Let's complete the implementation of the `serialization` and `deserialization`. FML provides the handle type `ser` and `de ` and resultant type is Bytes for the `SubStorageAccess`. Now, you need to implement the module level `ser` and `de` as you want.

basic_module/account/src/impls.rs
HoOngEe: We may need it in the future after host provides a suitable handling and a module decided to get help from the functionality.

basic_module/account/src/types.rs
```diff
@@ -14,11 +14,8 @@ pub struct Account {
 // Temporary measures for build.
 // This function will be removed when the structure of the context and conversion logic is completed.
```
HoOngEe, line 15: Here we should remove the command also.

basic_module/account/src/types.rs
```diff
@@ -87,3 +82,21 @@ impl Action {
         unimplemented!()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test() {
```
HoOngEe, line 91: We cannot test anything with this test because it requires a help from a human vision. It's better to make an assertion.

basic_module/account/src/types.rs
HoOngEe: Ah, Ok.

basic_module/account/src/impls.rs
```diff
@@ -36,3 +37,32 @@ impl CheckTxHandler for Handler {
...
+impl TransactionExecutor for Executor {
+    fn execute_transactions(
+        &self,
+        transactions: &[SignedTransaction],
+    ) -> Result<Vec<TransactionExecutionOutcome>, ()> {
+        let mut total_additional_fee: u64 = 0;
+        let mut total_min_fee: u64 = 0;
+
+        for signed_tx in transactions {
+            total_additional_fee += signed_tx.tx.fee;
```
HoOngEe, line 53: maybe the total_additional_fee should be signed_tx.tx.fee - signed_tx.tx.action.min_fee()

somniumism: Oh, I forgot it. I just modified it. Thanks

PR #289, opened 2020-03-26
[WIP] Implement Foundry Module Library a.k.a FML

junha1:
You don't have to look at the code unless you can give some advise about high-level interface design, before I remove WIP tag


basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+        panic!("Module hasn't terminated itself. Protocol Error")
+    }
+}
+
+/// This is for the SameProcess tasks. You can register a runnable function,
+/// and it will be executed later as an instance of 'process' (which is actually not).
+pub type ThreadAsProcesss = Arc<dyn Fn(Vec<String>) -> () + Send + Sync>;
+lazy_static! {
```
dynaxis, line 60: I recommend you to use [`once_cell`](https://crates.io/crates/once_cell) instead of `lazy_static`. It's currently in the process of being accepted into the `std`. It doesn't use macros at all.

junha1: Good to know. Thanks

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    fn join(&mut self) {
+        // PlainThread Executor is for test, so no worry for malicous unresponsiveness
+        self.handle.take().unwrap().join().unwrap();
+    }
+}
+
+/// Rust doesn't allow Drop for trait, so we need this. See E0120
+struct ExecutorWrapper<T: Executor> {
```
dynaxis, line 94: Why do we need a wrapper here?
```rust
trait Executor {
    fn join(&self) {}
}


struct A;

impl Drop for A {
    fn drop(&mut self) {
        self.join();
    }
}

impl Executor for A {
    fn join(&self) {}
}
```

???

junha1: @dynaxis I have to implement `Drop` for N time, for `struct A`, `struct B`, `struct C`, ...

dynaxis, line 94: Initially I misunderstood what you meant by the wrapper. If your intention is so, I think it's better named `ExecutorDropper` or similar. I may need to consult naming convention before coming up with a better name.

basesandbox/src/ipc/semaphore.rs
```diff
@@ -0,0 +1,125 @@
...
+            raw: create(&address, mode),
+            mode,
+            name: address,
+        }
+    }
+
+    fn ready(&mut self) {}
+}
+
+impl Semaphore {
```
dynaxis, line 102: To me, it's usually not good to retain codes without actual use elsewhere. If this is a generic library, then it's ok. Otherwise, I think it's better to remove it now and readd to when strictly required.

basesandbox/src/execution/executor.rs
junha1: Ok. That sounds better

junha1:
This belongs to https://github.com/CodeChain-io/baselink-rs

PR #290, opened 2020-03-27
Implement MemPoolAccess as a part of the context

HoOngEe:
~~It can only be merged after @Byeongjee 's branch `byeongjee/implement.validator.interface.call` is landed.
This pr includes only one commit `80bdd5`~~

### DIFF
--- a/coordinator/src/context.rs
+++ b/coordinator/src/context.rs
@@ -14,13 +14,16 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+use crate::Transaction;
 pub use ctypes::StorageId;
 
-/// A `Context` provides the interface against the system services such as DB access,
-pub trait Context: SubStorageAccess {}
+/// A `Context` provides the interface against the system services such as moulde substorage access,
+/// mempool access
+pub trait Context: SubStorageAccess + MemPoolAccess {}
 
 pub type Key = dyn AsRef<[u8]>;
 pub type Value = Vec<u8>;
+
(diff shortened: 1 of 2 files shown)


HoOngEe:
This work requires the type change of the argument `transaction` in the method `Client::queue_own_transaction`

HoOngEe:
I think that's because I had worked on the old branch, I'll update the pr.

HoOngEe:
@Byeongjee Ok, we firstly have to list up the errors that can be handled by modules. I think it's better to make a crate for such shared types.

Byeongjee:
> @Byeongjee Ok, we firstly have to list up the errors that can be handled by modules. I think it's better to make a crate for such shared types.

Maybe we can reuse `codechain-types` crate.

Byeongjee:
@HoOngEe Currently, the only error types currently available are
1.  `HistoryError::TransactionAlreadyImported`
2. `HistoryError::LimitReached`
3. `AppError(err_code)` (where `err_code` is a typical integer)

Note that `TransactionAlreadyImported` can be emitted in two cases
1. The transaction is already imported in the chain
2. The transaction is already imported in the mempool, but not included in a block yet.

`AppError` will be responsible for handling all other kinds of errors, including `HistoryError::Old` and `HistoryError::TooCheapToReplace`, etc.

Note that the error code of `AppError` is defined by modules, so the modules should know about the meaning of each error code, and are responsible for handling them.

If you think it's better to merge the error types into one, please let me know.

Byeongjee:
@HoOngEe How do you think about closing this PR now, and implement new error types, when we are actually implementing the error handling logic?

PR #291, opened 2020-03-28
Introduce DoubleVoteHandler

sgkim126:
This patch removes stake related implementations from the core crate.
The jobs that StakeHandler did are implemented in the state crate. The only thing that the state crate cannot handle is `ReportDoubleVote` because it's hard coupled with the consensus. So this patch introduces `DoubleVoteHandler`. I think a similar handler will be used even after the module separation.


sgkim126:
@majecty The first two commits are included in https://github.com/CodeChain-io/foundry/pull/283; please review other commits.

PR #300, opened 2020-04-03
Remove genesis_common_params from machine

Byeongjee:
Since CommonParams is saved to and restored from the state, the machine does not need to hold it.
FIXME in ready_transactions is resolved accordingly.

Now CodeChainMachine has no field, so I'm not sure if we further refactor this or not.

### DIFF
--- a/core/src/miner/mod.rs
+++ b/core/src/miner/mod.rs
@@ -103,7 +103,7 @@ pub trait MinerService: Send + Sync {
     ) -> Result<(TxHash, u64), Error>;
 
     /// Get a list of all pending transactions in the mem pool.
-    fn ready_transactions(&self, range: Range<u64>) -> PendingVerifiedTransactions;
+    fn ready_transactions(&self, size_limit: usize, range: Range<u64>) -> PendingVerifiedTransactions;
 
     /// Get list of all future transaction in the mem pool.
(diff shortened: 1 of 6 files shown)


core/src/miner/mod.rs
```diff
@@ -103,7 +103,7 @@ pub trait MinerService: Send + Sync {
     ) -> Result<(TxHash, u64), Error>;
 
     /// Get a list of all pending transactions in the mem pool.
-    fn ready_transactions(&self, range: Range<u64>) -> PendingVerifiedTransactions;
+    fn ready_transactions<C: EngineInfo>(&self, client: &C, range: Range<u64>) -> PendingVerifiedTransactions;
```
sgkim126, line 106: How about passing the max body size instead of passing EngineInfo?

Byeongjee: There will be a concept of `gas` in mold.
So in the future, gas limit parameter will be passed as well.
Do you think passing the two parameters will be better than passing `EngineInfo`?

Byeongjee: Ok

Byeongjee:
@sgkim126 Fixed.Please review again
I changed `CodeChainMachine::new()` to `Default::default()`.
CodeChainMachine will be removed when we implement the module system

PR #301, opened 2020-04-04
Remove CodeChainMachine

sgkim126:
CodeChainMachine was designed to abstract the rule of transaction and
state change, but it will be moved to the module. So this abstraction in
the core module is not necessary anymore.

There are five methods in CodeChainMachine.
- balance, add_balance, increase_term_id are not used.
- verify_transaction does nothing.
The only meaningful method is verify_transaction_seal. This commit makes
the caller call the internal code directly.


core/src/miner/miner.rs
```diff
@@ -167,7 +166,6 @@ impl Miner {
         mem_pool: &mut MemPool,
     ) -> Vec<Result<TransactionImportResult, Error>> {
         let best_header = client.best_block_header().decode();
-        let fake_header = best_header.generate_child();
```
majecty, line 170: Wow the `fake_header` was not used!

PR #305, opened 2020-04-06
Implement IPC

junha1:
This is part of https://github.com/CodeChain-io/foundry/pull/289 and created to merge it partially.

### DIFF
--- a/basesandbox/src/execution/executee.rs
+++ b/basesandbox/src/execution/executee.rs
@@ -14,5 +14,25 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pub mod execution;
-pub mod ipc;
+use crate::ipc::*;
+
+// Interface for the sandboxee written in Rust
+pub struct Context<T: Ipc> {
+    /// ipc will be given with Some, but module may take it
+    /// However, the ipc must be return back here before the module terminates
+    pub ipc: Option<T>,
+}
+
+pub fn start<T: Ipc>(mut args: Vec<String>) -> Context<T> {
+    let ipc = T::new(hex::decode(args.remove(1)).unwrap());
+    ipc.send(b"#INIT\0");
+    Context {
+        ipc: Some(ipc),
+    }
+}
+
+impl<T: Ipc> Context<T> {
+    pub fn terminate(self) {
+        self.ipc.unwrap().send(b"#TERMINATE\0");
+    }
+}
--- a/basesandbox/src/execution/executor.rs
+++ b/basesandbox/src/execution/executor.rs
@@ -1,0 +1,135 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::ipc::*;
+use once_cell::sync::OnceCell;
+use std::collections::HashMap;
+use std::process::Command;
+use std::sync::{Arc, RwLock};
+use std::time::Duration;
+
+pub trait Executor {
+    fn new(path: &str, args: &[&str]) -> Self;
+    fn join(&mut self);
+}
+
+pub struct Executable {
+    child: std::process::Child,
+}
+
+impl Executor for Executable {
+    fn new(path: &str, args: &[&str]) -> Self {
+        let mut command = Command::new(path);
+        command.args(args);
+        Executable {
+            child: command.spawn().unwrap(),
+        }
+    }
+
+    fn join(&mut self) {
+        // This is synchronized with excutee's signal (#TERMINATE),
+        // which is supposed to be sent in termination step.
+        // Thus, in normal case, it won't take much time to be in a waitable status.
+        // However a malicous excutee might send a signal arbitrarily before the termination.
+        // For that, we have a timeout for the wait.
+        for _ in 0..10 {
+            std::thread::sleep(std::time::Duration::from_millis(50));
+            if self.child.try_wait().unwrap().is_some() {
+                return // successful
+            }
+        }
+        panic!("Module hasn't terminated itself. Protocol Error")
+    }
+}
+
+/// This is for the intra-process tasks. You can register a runnable function,
+/// and it will be executed later as an instance of 'process' (which is actually not).
+pub type ThreadAsProcesss = Arc<dyn Fn(Vec<String>) -> () + Send + Sync>;
+
+static POOL: OnceCell<RwLock<HashMap<String, ThreadAsProcesss>>> = OnceCell::new();
+
+fn get_function_pool(key: &str) -> ThreadAsProcesss {
+    POOL.get_or_init(|| RwLock::new(HashMap::new())).read().unwrap().get(key).unwrap().clone()
+}
+
+pub fn add_function_pool(key: String, f: ThreadAsProcesss) {
+    POOL.get_or_init(|| RwLock::new(HashMap::new())).write().unwrap().insert(key, f);
+}
+
+pub struct PlainThread {
+    handle: Option<std::thread::JoinHandle<()>>,
+}
+
+impl Executor for PlainThread {
+    fn new(path: &str, args: &[&str]) -> Self {
+        let path = path.to_owned();
+        let mut args: Vec<String> = args.iter().map(|&x| x.to_string()).collect();
+        args.insert(0, "Thread".to_owned()); // corresponding to program path
+        let handle = std::thread::spawn(move || get_function_pool(&path)(args));
+
+        PlainThread {
+            handle: Some(handle),
+        }
+    }
+
+    fn join(&mut self) {
+        // PlainThread Executor is for test, so no worry for malicous unresponsiveness
+        self.handle.take().unwrap().join().unwrap();
+    }
+}
+
+/// Rust doesn't allow Drop for trait, so we need this. See E0120
+struct ExecutorDropper<T: Executor> {
+    executor: T,
+}
+
+impl<T: Executor> Drop for ExecutorDropper<T> {
+    fn drop(&mut self) {
+        self.executor.join();
+    }
+}
+
+/// declaration order of fields is important because of Drop dependencies
+pub struct Context<T: Ipc, E: Executor> {
+    pub ipc: T,
+    _child: ExecutorDropper<E>,
+}
+
+/// id must be unique for each instance.
+pub fn execute<T: Ipc, E: Executor>(path: &str) -> Result<Context<T, E>, String> {
+    let (config_server, config_client) = T::arguments_for_both_ends();
+    let ipc = T::new(config_server);
+    let config_client = hex::encode(&config_client);
+    let args: Vec<&str> = vec![&config_client];
+    let child = ExecutorDropper {
+        executor: Executor::new(path, &args),
+    };
+    let ping = ipc.recv(Some(Duration::from_millis(200))).unwrap();
+    assert_eq!(ping, b"#INIT\0");
+    Ok(Context {
+        ipc,
+        _child: child,
+    })
+}
+
+impl<T: Ipc, E: Executor> Context<T, E> {
+    /// Call this when you're sure that the excutee is ready to teminate; i.e.
+    /// it will call excutee::terminate() asap.
+    pub fn terminate(self) {
+        let ping = self.ipc.recv(Some(Duration::from_millis(200))).unwrap();
+        assert_eq!(ping, b"#TERMINATE\0");
+    }
+}
--- a/basesandbox/src/ipc.rs
+++ b/basesandbox/src/ipc.rs
@@ -1,0 +1,66 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pub trait IpcSend: Send {
+    /// It might block until counterparty's recv(). Even if not, the order is still guaranteed.
+    fn send(&self, data: &[u8]);
+}
+
+#[derive(Debug, PartialEq)]
+pub enum RecvError {
+    TimeOut,
+    Termination,
+}
+
+pub trait Terminate: Send {
+    /// Wake up block on recv with a special flag
+    fn terminate(&self);
+}
+
+pub trait IpcRecv: Send {
+    type Terminator: Terminate;
+
+    /// Returns Err only for the timeout or termination wake-up(otherwise panic)
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvError>;
+    /// Create a terminate switch that can be sent to another thread
+    fn create_terminator(&self) -> Self::Terminator;
+}
+
+pub trait Ipc: IpcSend + IpcRecv {
+    /// Generate two configurations
+    /// which will be feeded to Ipc::new(),
+    /// for both two ends in two different processes, to initialize each IPC end.
+    fn arguments_for_both_ends() -> (Vec<u8>, Vec<u8>);
+
+    type SendHalf: IpcSend;
+    type RecvHalf: IpcRecv;
+
+    /// Constructs itself with an opaque data that would have been transported by some IPC
+    fn new(data: Vec<u8>) -> Self;
+    /// split itself into Send-only and Recv-only. This is helpful for a threading
+    /// When you design both halves, you might consider who's in charge of cleaning up things.
+    /// Common implementation is making both to have Arc<SomethingDroppable>.
+    fn split(self) -> (Self::SendHalf, Self::RecvHalf);
+}
+
+/// Most of IPC depends on a system-wide name, which looks quite vulnerable for
+/// possible attack. Rather, generating a random name would be more secure.
+pub fn generate_random_name() -> String {
+    let pid = std::process::id();
+    let time = std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap();
+    let hash = ccrypto::blake256(format!("{:?}{}", time, pid));
+    format!("{:?}", hash)[0..16].to_string()
+}
--- a/basesandbox/src/ipc.rs
+++ b/basesandbox/src/ipc.rs
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 pub mod domain_socket;
+pub mod intra;
 
 pub trait IpcSend: Send {
     /// It might block until counterparty's recv(). Even if not, the order is still guaranteed.
--- a/basesandbox/src/ipc.rs
+++ b/basesandbox/src/ipc.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+pub mod domain_socket;
+
 pub trait IpcSend: Send {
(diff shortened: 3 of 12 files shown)


basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,166 @@
...
+    socket: Arc<SocketInternal>,
+    buffer: RefCell<Vec<u8>>,
+}
+
+impl IpcRecv for DomainSocketRecv {
+    type MyTerminate = Terminator;
+
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvFlag> {
+        self.socket.0.set_read_timeout(timeout).unwrap();
+        let (count, address) = self.socket.0.recv_from(&mut self.buffer.borrow_mut()).unwrap();
```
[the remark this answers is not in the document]
junha1: Well I will eventually replace it by stream (of course with buffering). I'm using datagram since it's easy to implement for now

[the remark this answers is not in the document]
junha1: If then arbitrary size of data will be transferable.

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,166 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+use std::cell::RefCell;
+use std::os::unix::net::UnixDatagram;
+use std::path::Path;
+use std::sync::Arc;
+
+const TEMPORARY_PATH: &str = "/tmp";
```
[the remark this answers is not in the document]
junha1: Good to know. Thanks

basesandbox/src/execution/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 Kodebox, Inc.
```
dynaxis, line 1: Though current Foundry codebase is largely following pre-2018 edition convention, it seems better to shy away from the use of `mod.rs`. In this case, move `mod.rs` up to the parent directory and rename it to `execution.rs`.

https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs

junha1: Wow. I really didn't know that

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pub mod domain_socket;
+pub mod intra;
+pub mod semaphore;
+use std::collections::hash_map::HashMap;
+
+/// Inter-process obejct that is constructed / synchronized by another Inter-process mechanism.
```
dynaxis, line 22: The overall naming and terms used in the comments disturb me. Inter-process itself means nothing about communication. So if you mean IPC, then use IPC. Not just `inter-process`.

In addition to the above point, there is a typo. ;) **obejct** --> object.

junha1: `InterProcessUnit` itself has nothing to do with IP'C'. It's just inter-processely initializable object. IPC endpoint is one possible implementor of `InterProcessUnit`

basesandbox/src/execution/executor.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    fn join(&mut self) {
+        // PlainThread Executor is for test, so no worry for malicous unresponsiveness
+        self.handle.take().unwrap().join().unwrap();
+    }
+}
+
+/// Rust doesn't allow Drop for trait, so we need this. See E0120
+struct ExecutorWrapper<T: Executor> {
```
dynaxis, line 95: As I suggested elsewhere, this is better named, for instance, `ExecutorCleaner` or `DropExecutor`. `Wrapper` doesn't manifest the sole role of this wrapper.

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.recv.recv(timeout)
+    }
+
+    fn create_terminate(&self) -> Self::MyTerminate {
+        self.recv.create_terminate()
+    }
+}
+
+impl Ipc for DomainSocket {
+    type SendOnly = DomainSocketSend;
```
dynaxis, line 156: I would have named them as `SenderHalf` and `ReceiverHalf`.

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    format!("{:?}", hash)[0..16].to_string()
+}
+
+pub trait IpcSend: Send {
+    /// It might block until counterparty's recv(). Even if not, the order is still guaranteed.
+    fn send(&self, data: &[u8]);
+}
+
+#[derive(Debug, PartialEq)]
+pub enum RecvFlag {
```
dynaxis, line 64: Isn't a `RecvError` better name?

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+pub struct DomainSocketRecv {
+    address_dst: String,
+    socket: Arc<SocketInternal>,
+    buffer: RefCell<Vec<u8>>,
+}
+
+impl IpcRecv for DomainSocketRecv {
+    type MyTerminate = Terminator;
```
dynaxis, line 84: `My` prefix should be avoided outside examples, IMHO. It doesn't convey meaningful information. Call it `Terminate` or `TerminateRecv`.

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvFlag>;
+    /// Create a terminate switch that can be sent to another thread
+    fn create_terminate(&self) -> Self::MyTerminate;
+}
+
+pub trait Ipc: InterProcessUnit + IpcSend + IpcRecv {
+    type SendOnly: IpcSend;
+    type RecvOnly: IpcRecv;
+    // split itself into Send-only and Recv-only. This is helpful for a threading
+    fn split(self) -> (Self::SendOnly, Self::RecvOnly);
```
dynaxis, line 87: If all `Ipc`s are always splitable into their sender half and receiver half, then it seems more natural just to have getters for them. Also it seems natural to return `Arc`. If not, it's conceptually confusing since an IPC primitive is likely to have links to the halves.

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pub mod domain_socket;
+pub mod intra;
+pub mod semaphore;
+use std::collections::hash_map::HashMap;
+
+/// Inter-process obejct that is constructed / synchronized by another Inter-process mechanism.
+pub trait InterProcessUnit {
```
dynaxis, line 23: Separation of `InterProcessUnit` trait seems overkill. Why a separate trait while `Ipc` seems enough? Too fine-grained abstraction might make codes difficult to grok.

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Constructs itself with an opaque data that would have been transported by some IPC
+    fn new(data: Vec<u8>) -> Self;
+}
+
+/// One possible abstraction for a server-client style of `Linker`
+pub trait TwoWayInitialize {
+    type Server: InterProcessUnit;
+    type Client: InterProcessUnit;
```
dynaxis, line 31: I can't find actual uses of `Server` and `Client`.

basesandbox/src/ipc/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+use std::collections::hash_map::HashMap;
+
+/// Inter-process obejct that is constructed / synchronized by another Inter-process mechanism.
+pub trait InterProcessUnit {
+    /// Constructs itself with an opaque data that would have been transported by some IPC
+    fn new(data: Vec<u8>) -> Self;
+}
+
+/// One possible abstraction for a server-client style of `Linker`
+pub trait TwoWayInitialize {
```
dynaxis, line 29: Do you anticipate any non `TwoWayInitialize` IPC? At least in the current context, this trait just seems to add complexity overall.

junha1: There could be some IPC scheme where client's init arguments can only be decided after the server's init. TwoWayInitialize doesn't support that.

basesandbox/src/ipc/mod.rs
junha1: The only example that is `InterProcessUnit` but not IPC was the semaphore.
Since I'll remove that, I'll remove this too.

basesandbox/src/ipc/mod.rs
junha1: What do you mean by `getter`? You mean re-combine?

basesandbox/src/ipc/mod.rs
junha1: Well `Termination` is not an error I think.

junha1: `Error` sounds to harsh for normal, expected path like `Termination`

basesandbox/src/ipc/domain_socket.rs
junha1: That's better. Thanks

basesandbox/src/execution/executor.rs
junha1: Got it

basesandbox/src/ipc/mod.rs
junha1: That will be `ThreeWayInitialize` maybe.

basesandbox/src/ipc/mod.rs
junha1: Ok

basesandbox/src/ipc/mod.rs
junha1: They are reserved for even higher abstraction that takes `T: TwoWayInitialize`, and perform generic way of utilizing it, possibly in the coordinator. Personally I enjoy this kind of further-use abstraction, but also agree that it has no reason to exit for now. I'll remove it after I check they are not actually used in FML part.

basesandbox/src/ipc/domain_socket.rs
junha1: Oh sure. My mistake.

basesandbox/src/ipc/mod.rs
junha1: Note that it takes `self`, not `&self`.

basesandbox/src/ipc/mod.rs
junha1: However, this is an exact example of case where you can give me a good advice, since you know much better about various IPCs than me. If you think there won't be any `ThreeWayInitialize` IPC, then please tell me so that I can remove this unnecessary abstraction

basesandbox/src/execution/mod.rs
junha1: @majecty What do you think of this? Should I apply this rule for my work? (Since my work is not related to codechain codebase, so might be ok to be try the latest convention?)

majecty: It seems good. Since many other modules in Foundry are using `mod.rs`, how about sharing this with all other team members?

basesandbox/src/ipc.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Constructs itself with an opaque data that would have been transported by some IPC
+    fn new(data: Vec<u8>) -> Self;
+    /// split itself into Send-only and Recv-only. This is helpful for a threading
+    fn split(self) -> (Self::SendHalf, Self::RecvHalf);
+}
+
+/// Some implementors of InterProcessUnit would take a single 'path'
+/// since it's quite a trivial model that inter-process objects try to acquire an access to a
+/// inter-process handle that is identified by a system-wide name.
+pub fn init_data_from_path(path: String) -> Vec<u8> {
```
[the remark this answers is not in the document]
majecty: I have a call with Junha. He will check the usage of this function.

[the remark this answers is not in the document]
junha1: It is not used. I'll remove it

basesandbox/src/ipc.rs
```diff
@@ -0,0 +1,80 @@
...
+    type Terminator: Terminate;
+
+    /// Returns Err only for the timeout or termination wake-up(otherwise panic)
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvError>;
+    /// Create a terminate switch that can be sent to another thread
+    fn create_terminate(&self) -> Self::Terminator;
+}
+
+/// One possible abstraction for a server-client style of `Linker`
+pub trait TwoWayInitialize {
```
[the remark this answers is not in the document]
majecty: I had a call with Junha. Twoway means that it will use two-way handshake.

basesandbox/src/execution/executee.rs
```diff
@@ -0,0 +1,38 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::ipc::*;
+
+// Interface for the sandboxee written in Rust
+pub struct Context<T: Ipc> {
+    /// ipc will be given with Some, but module may take it
+    /// However, the ipc must be return back here before the module terminates
+    pub ipc: Option<T>,
```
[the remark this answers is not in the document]
junha1: If the executee want to utilize given ipc with its own purpose, it will `take()` it. However the ipc has to be used in destruction phase, the excutee must return back it.

basesandbox/src/execution/executee.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+pub fn start<T: Ipc>(args: Vec<String>) -> Context<T> {
+    let ipc = T::new(hex::decode(args[1].clone()).unwrap());
```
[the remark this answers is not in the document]
junha1: You're right. Thanks

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,156 @@
...
+            serde_cbor::to_vec(&(&self.address_server, &self.address_client)).unwrap(),
+            serde_cbor::to_vec(&(&self.address_client, &self.address_server)).unwrap(),
+        )
+    }
+}
+
+impl Drop for DomainSocketLinker {
+    fn drop(&mut self) {
+        std::fs::remove_file(&self.address_server).unwrap();
+        std::fs::remove_file(&self.address_client).unwrap();
```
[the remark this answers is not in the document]
majecty: I had a call with Junha. He will move the remove code to `DomainSocket`.

basesandbox/src/ipc.rs
```diff
@@ -0,0 +1,80 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use std::collections::hash_map::HashMap;
+
+pub trait IpcSend: Send {
```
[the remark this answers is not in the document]
majecty: The types will be used in the next PR.

basesandbox/src/ipc/domain_socket.rs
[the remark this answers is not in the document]
junha1: @majecty I tested this `assert!(count <= self.buffer.borrow().len(), "Unix datagram got data larger than the buffer.");` What do you mean besides this?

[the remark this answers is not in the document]
majecty: Our current implementation does not handle a packet whose size is greater than 1024.
Since the current implementation returns first 1024 byte. It may make the deserializer fails.
Instead of returning the first 1024 bytes of packet data, creating panic with explicit reason will make debugging easier.

```rust
if (count == self.buffer.borrow().len()) {
  panic!("Our unix datagram implementation does not support packet whose length is greater than 1024.");
}
```

basesandbox/src/execution/executee.rs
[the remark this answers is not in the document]
junha1: This is because I assumed `Ipc` can be not `Sync`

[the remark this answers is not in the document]
majecty: It seems awkward. Let's talk again when actually the thread takes the ipc field.

basesandbox/src/ipc/domain_socket.rs
```diff
@@ -0,0 +1,151 @@
...
+    buffer: RefCell<Vec<u8>>,
+}
+
+impl IpcRecv for DomainSocketRecv {
+    type Terminator = Terminator;
+
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvError> {
+        self.socket.0.set_read_timeout(timeout).unwrap();
+        let (count, address) = self.socket.0.recv_from(&mut self.buffer.borrow_mut()).unwrap();
+        assert!(count < self.buffer.borrow().len(), "Unix datagram got data larger than the buffer.");
```
majecty, line 79: :+1:

basesandbox/src/ipc.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn terminate(&self);
+}
+
+pub trait IpcRecv: Send {
+    type Terminator: Terminate;
+
+    /// Returns Err only for the timeout or termination wake-up(otherwise panic)
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvError>;
+    /// Create a terminate switch that can be sent to another thread
+    fn create_terminate(&self) -> Self::Terminator;
```
dynaxis, line 42: It returns a `Terminator` but named `create_terminate`.

basesandbox/src/ipc.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn new(name: String) -> Self;
+    /// Generate two configurations
+    /// which will be feeded to Ipc::new(),
+    /// for both two ends in two different processes.
+    fn create(&self) -> (Vec<u8>, Vec<u8>);
+}
+
+// TODO: Rename this with TwoWayInitializableIpc when we actually add ThreeWayInitialize or so.
+pub trait Ipc: IpcSend + IpcRecv {
+    type Linker: TwoWayInitialize;
```
dynaxis, line 59: `Linker` is not referenced elsewhere. Why leave it here?

basesandbox/src/ipc.rs
junha1: Ok

basesandbox/src/ipc.rs
junha1: @dynaxis It is used here https://github.com/CodeChain-io/foundry/pull/305/files#diff-c49b0dbd7a77e7b5b4ca2550156bfa28R114

basesandbox/src/ipc.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn recv(&self, timeout: Option<std::time::Duration>) -> Result<Vec<u8>, RecvError>;
+    /// Create a terminate switch that can be sent to another thread
+    fn create_terminate(&self) -> Self::Terminator;
+}
+
+/// One possible abstraction for a server-client style of `Linker`
+/// TwoWay means you there are exactly two communcations to establish IPC.
+/// And moreover, those two communications have no dependencies, so it trivially returns
+/// two arguments (to be sent).
+pub trait TwoWayInitialize {
```
dynaxis, line 49: I'm still not sure if `TwoWayInitialize` is a necessary abstraction here. It just produces a pair of messages and there is no direct relationship with `Ipc`.

junha1: Could you please take a look of this https://github.com/CodeChain-io/foundry/pull/305#discussion_r409568704

dynaxis: If `Linker` is used that way, I still don't get why `TwoWayInitialize` should be separate. Do you have any strong requirement to reuse single impl of `TwoWayInitialize` for multiple different `Ipc`s?

junha1: @dynaxis I don't get what you mean. `TwoWayInitialize` is implemented at least once for every IPCs.

dynaxis: 1. `TwoWayInitialize` just generates a pair of arguments. So see no reason why it should be a separate `object` at all.
1. `TwoWayInitialize` is related to an IPC mechanism. And therefore implemented for **each IPC**. Then is there any reason to not merge it into `Ipc`? It's anyway specific to an IPC mechanism. Note that you can have trait level functions without taking a `self`.
1. If it's ever important to have a linker separate (in your design I mean. The linker in my API design is a bit different), I would rather have made the linking process more into the linker concept, rather than keeping a separate function such as `execute`. Then you can hide more from the outside.

The above are all of my points.

junha1: @dynaxis At first in the PR, the `Linker` is assumed to be object so that it can stay alive after linkage, for cleaning up purpose. (current generic function of `execute` can take advantage of possible `linker`'s `Drop`) And there was actually such `Drop` impl for `DomainSocket`, but it was turned out to be ok to be just in each end, as @majecty requested. So yes, you're right on this point. For current two implementations of `Ipc`, it is ok with `TwoWayInitialize` compensated with a single static (no `&self`) function. If this bothers you, I'll just apply your suggestion.

PR #306, opened 2020-04-07
Using tokio make unrelated test fail

majecty:
network2.test.ts failed when we add `tokio` as a dependency.

## How to reproduce:

Run the command below in the test directory.
```
cargo build && yarn mocha -r ts-node/register --timeout 5000 src/e2e/network2.test.ts
```


majecty:
I created this PR to share the failed tests with others.
Since I fixed it at https://github.com/CodeChain-io/foundry/pull/309, I'll close this PR.

Issue #308, opened 2020-04-07
Cannot receive HUP event from mio >= 0.6.17


majecty:
In the current master (45dcf9d27b1339f940bd98bb31676f26fdb38146) branch, if we upgrade the mio library, Foundry can't receive sockets' `HUP` events.

You can easily check it by upgrading the mio version and running the command below:
```
yarn mocha -r ts-node/register --timeout 5000 src/e2e/network2.test.ts
```

I found that mio's this commit changed the behavior: https://github.com/tokio-rs/mio/commit/b89e3bd662c27242909b446dbb80c74249fe6042

Using other libraries that have a mio >= 0.6.17 as a dependency (tokio is the example) causes the problem.

~We are using the deprecated APIs of the mio. I suspect that the cause of the problem came from the usage of the deprecated APIs.~

I found that we should not depend on the "HUP" event in the mio document.
> In other words, portable programs that explicitly check for [hup](https://docs.rs/mio/0.6.21/mio/unix/struct.UnixReady.html#method.hup) or [error](https://docs.rs/mio/0.6.21/mio/unix/struct.UnixReady.html#method.error) readiness should be doing so as an **optimization** and always be able to handle an error or HUP situation when performing the actual read operation.
https://docs.rs/mio/0.6.21/mio/struct.Poll.html#readiness-operations

I also checked that the return value of the read is zero when a connection is canceled.
> If n is 0, then it can indicate one of two scenarios: 
This reader has reached its "end of file" and will likely no longer be able to produce bytes. Note that this does not mean that the reader will __always__ no longer be able to produce bytes.
https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read

However, our code depends on the HUP event from the mio. We should check the return value of the read.

PR #309, opened 2020-04-08
Disconnect a socket If Foundry meets EOF while reading the socket

majecty:
This PR fixes https://github.com/CodeChain-io/foundry/issues/308

There are a few things to do more.

1. Should we catch the error in the places?
2. What should we do when we met other IO errors? Should we disconnect the socket?
3. What happens if we write bytes to a disconnected socket? (This PR only checks the `read` call).

### DIFF
--- a/network/src/p2p/handler.rs
+++ b/network/src/p2p/handler.rs
@@ -941,32 +941,56 @@ impl IoHandler<Message> for Handler {
         Ok(())
     }
 
-    fn stream_writable(&self, _io: &IoContext<Message>, stream: StreamToken) -> IoHandlerResult<()> {
+    fn stream_writable(&self, io: &IoContext<Message>, stream: StreamToken) -> IoHandlerResult<()> {
         match stream {
             FIRST_INBOUND..=LAST_INBOUND => {
                 if let Some(con) = self.inbound_connections.write().get_mut(&stream) {
-                    con.flush()?;
+                    let flush_result = con.flush();
+                    if let Err(P2PConnectionError::IoError(io_error)) = &flush_result {
+                        if io_error.kind() == std::io::ErrorKind::BrokenPipe {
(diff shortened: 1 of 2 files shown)


test/src/e2e/network2.test.ts
```diff
@@ -27,6 +27,8 @@ describe("network2 nodes", function() {
         nodeA = new CodeChain();
         nodeB = new CodeChain();
         await Promise.all([nodeA.start(), nodeB.start()]);
+        nodeA.keepLogs();
```
sgkim126, line 30: Should we always keep the logs?
As far as I remember, it makes the Travis upload these logs even though this test succeeds when other tests in the e2e directory fail.

majecty: I created the second commit to test this PR. I'll remove this commit later.
https://github.com/CodeChain-io/foundry/pull/309/commits/31225289d5f6cdfd9208889722cf7a8d9dc14467

majecty:
@sgkim126 Thanks I'll develop this PR.

I found these sentences from the Write document.
> A return value of 0 typically means that the underlying object is no longer able to accept bytes and will likely not be able to in the future as well, or that the buffer provided is empty.
How about disconnecting the socket if the `write` function returns 0?

https://doc.rust-lang.org/std/io/trait.Write.html

majecty:
I found this article from ibm.com
https://developer.ibm.com/technologies/linux/tutorials/l-sockpit/

> The closure of a peer socket can also be detected with the write API function. In this case, you’ll receive a SIGPIPE signal or, if this signal is blocked, the write function will return a -1 and set errno to EPIPE.

majecty:
@MSNTCS This RP fixes the mio issue.

majecty:
@sgkim126 I'm ready for you review.

PR #318, opened 2020-04-10
Implement Orderer module

Byeongjee:
Orderer module will have a single method `filter_and_order_transactions`.


basic_module/orderer/src/core.rs
```diff
@@ -0,0 +1,21 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use super::types::{OrdererError, TxMetadata};
+
+pub trait Orderer {
+    fn filter_and_order_transactions(metadata_list: Vec<TxMetadata>) -> Result<Vec<usize>, OrdererError>;
```
junha1, line 20: &self

Byeongjee: So, each handle must have a dummy trait and a dummy self reference, even if they are not used?

junha1: What do you mean by 'dummy trait'?

Byeongjee: The current Orderer trait is a dummy trait. 
It doesn't use `self`, so it doesn't have to be a trait at all.

If there is some way of registering a global function as a callback, we don't need the Orderer module at all.

Byeongjee:
@dynaxis @somniumism @HoOngEe 
Since implementing the orderer does not require a lot of work, I think it’s better to finish the orderer after fixing all specifications of the coordinator-orderer interface, and distribution module requirements. Implementing the orderer won't be a bottleneck.

PR #320, opened 2020-04-11
No need to import a macro explicitly

sgkim126:


sgkim126:
@Byeongjee I changed the commit message.

PR #321, opened 2020-04-11
Change the fields of ConsensusParams to non optional

sgkim126:
Do you have a reason to make them optional?


Byeongjee:
@sgkim126 I think you forgot to add me as a reviewer.
No, it was just a mistake to just copy them from `json/src/scheme/tendermint.rs`
But anyway it is resolved in my pr #260. I introduced `ConsensusParams` type with non-optional parameters.

PR #326, opened 2020-04-13
Verify headers in header sync

majecty:


types/src/sync_header.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    pub fn number(&self) -> BlockNumber {
+        self.block_header.number()
+    }
+
+    pub fn inner(&self) -> &Header {
+        &self.block_header
+    }
+
+    pub fn into_inner(self) -> Header {
```
sgkim126, line 49: Why not implement `From` instead of the custom method?

types/src/validator_set.rs
```diff
@@ -36,6 +36,10 @@ impl CompactValidatorSet {
     pub fn hash(&self) -> H256 {
         blake256(self.rlp_bytes())
     }
+
+    pub fn into_inner(self) -> Vec<CompactValidatorEntry> {
```
sgkim126, line 40: ditto

types/src/sync_header.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    pub fn hash(&self) -> BlockHash {
+        self.block_header.hash()
+    }
+
+    pub fn number(&self) -> BlockNumber {
+        self.block_header.number()
+    }
+
+    pub fn inner(&self) -> &Header {
```
sgkim126, line 45: And why not implement `Deref` for it?

majecty:
@sgkim126 I updated this PR.
I found that I didn't check the next_validator_set and hash of current_validator_set. I'll create a new PR to check them.

types/src/validator_set.rs
sgkim126: It seems hat it's not fixed yet.

majecty: Sorry, I missed it.
I fixed it now.

majecty:
@sgkim126 Could you check this PR again?

types/src/sync_header.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl SyncHeader {
+    pub fn new(block_header: Header, validator_set: Option<CompactValidatorSet>) -> Self {
+        Self {
+            block_header,
+            prev_validator_set: validator_set,
+        }
+    }
+
+    pub fn hash(&self) -> BlockHash {
```
sgkim126, line 38: Because the dot operator automatically calls deref, you don't need to implement `hash`, `number`, and `parent_hash`.

core/src/verification/queue/kind.rs
```diff
@@ -98,9 +98,9 @@ pub mod headers {
     pub struct Headers;
 
     impl Kind for Headers {
-        type Input = Header;
-        type Unverified = Header;
-        type Verified = Header;
+        type Input = SyncHeader;
+        type Unverified = SyncHeader;
+        type Verified = SyncHeader;
```
sgkim126, line 103: I think we don't need the validator set after the header is verified.
How about keeping `Verified` as `Header`?

majecty: In the next PR, the importer will use SyncHeader to check validator set hash.

The commit: https://github.com/CodeChain-io/foundry/pull/386/commits/1bf7c58edb5b11f17b939e8e56d8d53e3dc2fbf7
The PR: https://github.com/CodeChain-io/foundry/pull/386

types/src/sync_header.rs
majecty: I fixed it!

core/src/verification/queue/kind.rs
sgkim126: okay.

majecty:
Siince the failed test is "connect 100 nodes", I'll merge this PR.

PR #327, opened 2020-04-14
Remove deprecated comment about static validator

majecty:
We removed static validator set before.


junha1:
Related: https://github.com/CodeChain-io/foundry/issues/158

PR #328, opened 2020-04-14
[WIP] Implement Distribution module

somniumism:


basic_module/distribution/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "distribution"
+version = "0.1.0"
+authors = ["CodeChain Team <hi@codechain.io>"]
+edition = "2018"
+
+[dependencies]
+primitives = { git = "https://github.com/CodeChain-io/rust-codechain-primitives.git", version = "0.4" }
+coordinator = {package = "coordinator", path = "../../coordinator" }
+ckey = { package = "codechain-key", path = "../../key" }
```
HoOngEe, line 10: Add a new line please.

basic_module/distribution/src/core.rs
```diff
@@ -0,0 +1,7 @@
+pub trait Distribution {
+    fn distribute();
```
junha1, line 2: All methods must have `&self`

basic_module/distribution/src/core.rs
```diff
@@ -0,0 +1,7 @@
...
+}
+
+pub trait FeeManager {
+    fn recode_fee(fee: u64);
```
junha1, line 6: ditto

basic_module/distribution/src/core.rs
```diff
@@ -0,0 +1,7 @@
+pub trait Distribution {
+    fn distribute();
+}
+
+pub trait FeeManager {
```
HoOngEe, line 5: It seems like an exported trait for other modules to record fees, doesn't it? Then, the method name should be "record_fee"

basic_module/distribution/Cargo.toml
Byeongjee: @somniumism I recommend you to install EditorConfig for VSCode plugin.
I can help you if you have any problem while installing it.

basic_module/distribution/src/impls.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::collections::{BTreeMap, HashSet};
+
+#[allow(dead_code)]
+struct DistributionState {
+    reward: IntermediateRewards,
+
+    // `validators` is used for inactive validators after distributing rewards
+    inactive_validators: HashSet<Address>,
+
+    // `next_validator` is used to store updated validator set to TopLevelState
```
Byeongjee, line 13: Why Distribution module manages validators in its state?
I think it should be managed by Staking module.

basic_module/distribution/src/core.rs
```diff
@@ -0,0 +1,7 @@
+pub trait Distribution {
```
HoOngEe, line 1: Distribution module should record each validator's contribution when abci `BeginBlock` and distribute should be called when abci "EndBlock" if the module-defined term ends. `BeginBlockHandler` and `EndBlockHandler` can be more clear for those trait names.

basic_module/distribution/src/impls.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl Distribution for DistributionState {
+    fn distribute(&self) {}
+}
+
+#[allow(dead_code)]
+struct Manager {}
+
+impl FeeManager for Manager {
+    fn record_fee(&self, _fee: u64) {}
```
Byeongjee, line 39: What is the exact role of `record_fee`?
If you want to provide an interface to calculate each block fee by accumulation, I think the name should explicitly describe the behavior. Maybe something like `add_block_fee` or `accumulate_block_fee` work, but I'm not quite sure.

Also, I think you need hooks for `open_block` and `close_block`. (Maybe `close_block` can be merged into `distribute`)

basic_module/distribution/src/core.rs
Byeongjee: I prefer `open_block` and `close_block`. We use the names.

somniumism: How about these trait names `OpenBlockHandler` and `CloseBlockHandler`?

basic_module/distribution/src/impls.rs
somniumism: `accumulate_block_fee` looks good to me.

Byeongjee:
Why don't you use `Header `type defined in `codechain-types`?

basic_module/distribution/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+
+#[derive(Clone, Debug)]
+pub struct IntermediateRewards {
+    current_reward: BTreeMap<Public, u64>,
+    calculated_reward: BTreeMap<Public, u64>,
+    total_min_fee: u64,
+    min_fee: u64,
+}
+
+impl IntermediateRewards {
```
HoOngEe, line 52: Where is the function loading the state from the storage?

basic_module/distribution/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+}
+
+fn share(total_stakes: u64, stake: u64, total_min_fee: u64) -> u64 {
+    assert!(total_stakes >= stake);
+    u64::try_from((u128::from(total_min_fee) * u128::from(stake)) / u128::from(total_stakes)).unwrap()
+}
+
+pub struct ActionDataKeyBuilder {
```
HoOngEe, line 159: We don't need `ActionDataKeyBuilder` because when you request `set` to `substorageAccess` the key will be prefixed in the host side so that entries will not be overlapped with other modules.

basic_module/distribution/src/lib.rs
HoOngEe: Ah, ok the function is not implemented for the struct but for a global function

basic_module/distribution/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    pub fn update_quantity(&mut self, address: &Public, quantity: u64) {
+        if quantity == 0 {
+            return
+        }
+        *self.current_reward.entry(*address).or_insert(0) = quantity;
+    }
+
+    pub fn drain_calculated_rewards(&mut self) -> StateResult<BTreeMap<Public, u64>> {
```
HoOngEe, line 76: `StateResult` return type is weird, because the storage interface does not emit its error

basic_module/distribution/src/lib.rs
somniumism: Yes, I implemented `get_intermediate_rewards()` as a global function.

basic_module/distribution/src/lib.rs
somniumism: I removed `ActionDataKeyBuilder` and modified the login in `get_key()` to return `H256::default()`.

basic_module/distribution/src/lib.rs
somniumism: I removed `StateResult` and modified `pub fn drain_calculated_rewards(&mut self) -> StateResult<BTreeMap<Public, u64>>` to `-> BTreeMap<Public, u64>`

basic_module/distribution/src/lib.rs
HoOngEe: Instead of the `H256::default()` you can define any key value like "intermediate_reward" inside the module. The only required trait bound is `AsRef<[u8]>`.

basic_module/distribution/src/export.rs
```diff
@@ -0,0 +1,55 @@
+use ckey::Ed25519Public as Public;
```
HoOngEe, line 1: Maybe you wanted to write `import.rs`

basic_module/distribution/src/types.rs
```diff
@@ -13,3 +13,115 @@
...
+    },
+    OpenBlock {
+        block_number: u64,
+        block_author: Public,
+        block_hash: BlockHash,
+        rewards: BTreeMap<Public, u64>,
+    },
+}
+
+const MAX_VALIDATOR_SIZE: usize = 800;
```
HoOngEe, line 73: Currently, it is fine. Later, we're going to utilize `foundry / types` crate to share types between host and modules.

basic_module/distribution/src/types.rs
```diff
@@ -13,3 +13,115 @@
...
+const BITSET_SIZE: usize = MAX_VALIDATOR_SIZE / 8;
+
+#[derive(Copy, Clone)]
+pub struct BitSet([u8; BITSET_SIZE]);
```
HoOngEe, line 77: Currently, it is fine. Later, we're going to utilize foundry / types crate to share types between host and modules.

basic_module/distribution/src/types.rs
```diff
@@ -13,3 +13,115 @@
...
+    /// Network Id
+    pub network_id: NetworkId,
+
+    pub action: Action,
+}
+
+#[allow(dead_code)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Action {
+    CloseBlock {
```
HoOngEe, line 60: I think `Action` names are better to be descriptive such as `DistributeRewards`

basic_module/distribution/src/lib.rs
```diff
@@ -13,3 +13,305 @@
...
+    // This function should be implemented after the context has been formatted.
+    unimplemented!();
+}
+
+pub fn get_storage_id() -> StorageId {
+    // This function should be implemented after the storageId has been formatted.
+    unimplemented!()
+}
+
+pub fn get_key() -> H256 {
```
HoOngEe, line 176: You can name the key arbitrarily. Module level keys don't have to have `H256` types.

basic_module/distribution/src/lib.rs
```diff
@@ -13,3 +13,305 @@
...
+
+#[allow(dead_code)]
+#[derive(Default)]
+struct WorkInfo {
+    proposed: usize,
+    missed: usize,
+    signed: u64,
+}
+
+fn aggregate_work_info(start_of_the_next_term_header: Header) -> Result<HashMap<Public, WorkInfo>, Error> {
```
HoOngEe, line 225: I think this function can not expect any errors. Is the function signature intentional ?

basic_module/distribution/src/core.rs
```diff
@@ -13,3 +13,18 @@
...
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::error::Error;
+use ckey::Ed25519Public as Public;
+use ctypes::BlockHash;
+use std::collections::BTreeMap;
+
+pub trait OpenBlockHandler {
+    fn distribute_when_open_block(
```
HoOngEe, line 23: You should consult the abci interface functions. The function signature don't seem like this.

basic_module/distribution/src/impls.rs
```diff
@@ -13,3 +13,54 @@
...
+use std::collections::BTreeMap;
+
+#[allow(dead_code)]
+struct Distribution {
+    block_number: u64,
+    block_author: Public,
+}
+
+impl OpenBlockHandler for Distribution {
+    fn distribute_when_open_block(
```
HoOngEe, line 35: I think what you should do in `open_block` is recording some necessary information. You should execute distribution when executing the corresponding transaction.

HoOngEe: ditto in `close_block`

basic_module/distribution/src/lib.rs
```diff
@@ -58,7 +59,7 @@ impl IntermediateRewards {
 
         context.set(
             get_storage_id(),
-            &get_key(),
+            &StorageKey::default(),
```
HoOngEe, line 62: I think it's unnatural to utilize [0; 32] as a key for the entry `intermediate_reward`. As I [mentioned](https://github.com/CodeChain-io/foundry/pull/328#discussion_r414313470), just a simple descriptive key value is enough.

somniumism: So you mean, any type such as u16 or u64 is fine?

basic_module/distribution/src/core.rs
```diff
@@ -41,6 +27,12 @@ pub trait TransactionHandler {
     fn create_distribution_transaction(&self) -> Transaction;
 
     fn create_open_block_transaction(&self, header: &Header) -> Transaction;
+}
+
+pub trait BlockExecutor {
+    fn open_block(&self, context: &mut dyn SubStorageAccess, header: &Header, verified_crime: &[VerifiedCrime]);
+
+    fn execute_transactions(&self, context: &mut dyn SubStorageAccess, transactions: &[Transaction]);
```
Byeongjee, line 35: @HoOngEe @somniumism What was our conclusion on managing context?
Should `context` explicitly exposed to each module?

HoOngEe: Do you mean the explicit `context` parameter? I think it is desirable though currently not supported.

basic_module/distribution/src/impls.rs
```diff
@@ -14,108 +14,42 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use crate::core::{CloseBlockHandler, FeeManager, OpenBlockHandler, TransactionHandler};
+use super::core::*;
```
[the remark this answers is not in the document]
somniumism: It's a mistake. I fixed it. Thanks

basic_module/distribution/src/impls.rs
```diff
@@ -112,3 +112,71 @@ impl FeeManager for Distribution {
         context.set(get_storage_id(), &get_key(), intermediate_rewards.into());
     }
 }
+
+impl TransactionHandler for Distribution {
+    fn create_distribution_transaction(&self) -> Transaction {
```
HoOngEe, line 51: What do you think of changing the type of the transaction `DstributeRewards` more descriptive? The original intention of  introducing such explicit `DistributeRewards` transaction is to make state changes more visible. Though all state transition rules can not be expressed just by its type, we are able to improve it. If the transaction type contains only the map from accounts to distributed rewards then the state tracking can be more easier.
By complicating creation logic, you can simplify its execution.

basic_module/distribution/src/impls.rs
```diff
@@ -100,3 +100,15 @@ impl CloseBlockHandler for Distribution {
         Ok(())
     }
 }
+
+impl FeeManager for Distribution {
```
HoOngEe: I don't think the struct `Distribution` is not responsible for the `FeeManager` functionality. `Distribution` contains the internal state `block_number` and `author` to remember in what context it is executing transactions. However, `FeeManager` can achieve its goal regardless of the state of `Distribution`. In the sense that it will be instantiated and linked to another module, the variable in this instance will never be used. Therefore, it's better to implement `FeeManager` trait with another struct.

basic_module/distribution/src/impls.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn accumulate_block_fee(&self, total_additional_fee: u64, total_min_fee: u64) {
+        let context = get_context();
+        let mut intermediate_rewards: IntermediateRewards = get_intermediate_rewards(get_key());
+
+        intermediate_rewards.total_additional_fee += total_additional_fee;
+        intermediate_rewards.total_min_fee += total_min_fee;
+
+        context.set(get_storage_id(), &get_key(), intermediate_rewards.into());
```
Byeongjee: Why don't you use `intermediate_rewards.save_to_state()`?
One of the two code snippets should be changed.

somniumism:
TODO:
- [ ] Add logic that verify whether the reward amount is correct

PR #331, opened 2020-04-16
Remove unnecessary serialize/deserialize while calling import_header

majecty:
@sgkim126 I changed import_trusted_header to receive `Header` instead of `&Header`. When importing a header, I feel moving a header is more natural than borrowing the header. However, I'm not confident. Do you have any idea?


sgkim126:
@majecty I think it's good to move the value if there is no overhead.

majecty:
Since the failed tests are "discovery5 nodes" and "connect 100 nodes", I merged this PR.

PR #332, opened 2020-04-16
Rename import functions

majecty:


majecty:
Since the failed test is "connect 100 nodes", I'll merge this PR.

PR #333, opened 2020-04-18
Simplify miner, client and block.

sgkim126:
This refactoring is done to implement https://github.com/CodeChain-io/foundry/pull/324. But, because the PR takes more time than expected, I extracted the refactoring commits.


sgkim126:
The linux build fails, but the same test successes on osx. I think it's the problem of Travis.

PR #335, opened 2020-04-20
Implement Staking module

HoOngEe:
I tried to simulate staking logic previously is operating inside the host.
If it is merged into mold, the following tasks should be carried out.
1. Remove staking logic from the host.
2. Modify `TopLevelState` to store only CompactValidatorSet which contains only validators' public key and abstract voting powers. 
3. Remove parameters only used by `Staking`.

Followings are not implemented yet:
1. Executing `ReportDoubleVote` and banning malicious validators.
2. Creating transactions from higher level input (e.g. graphQL mutation request)

### DIFF
--- a/basic_module/staking/src/error.rs
+++ b/basic_module/staking/src/error.rs
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 pub use crate::runtime_error::Error as RuntimeError;
+pub use crate::syntax_error::Error as SyntaxError;
 use std::fmt::{Display, Formatter, Result as FormatResult};
 
 #[derive(Debug)]
@@ -49,10 +50,17 @@ impl<T: Display> Display for Insufficient<T> {
 #[derive(Debug)]
 pub enum Error {
     Runtime(RuntimeError),
+    Syntax(SyntaxError),
 }
 
 impl From<RuntimeError> for Error {
     fn from(error: RuntimeError) -> Error {
         Error::Runtime(error)
+    }
+}
+
+impl From<SyntaxError> for Error {
+    fn from(error: SyntaxError) -> Error {
+        Error::Syntax(error)
     }
 }
--- a/basic_module/staking/src/error.rs
+++ b/basic_module/staking/src/error.rs
@@ -1,0 +1,58 @@
+// Copyright 2020 Kodebox, Inc.
+// This file is part of CodeChain.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pub use crate::runtime_error::Error as RuntimeError;
+use std::fmt::{Display, Formatter, Result as FormatResult};
+
+#[derive(Debug)]
+/// Error indicating an expected value was not found.
+pub struct Mismatch<T> {
+    /// Value expected.
+    pub expected: T,
+    /// Value found.
+    pub found: T,
+}
+
+impl<T: Display> Display for Mismatch<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FormatResult {
+        write!(f, "Expected {}, found {}", self.expected, self.found)
+    }
+}
+
+#[derive(Debug)]
+pub struct Insufficient<T> {
+    /// Value to have at least
+    pub required: T,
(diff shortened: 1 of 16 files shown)


HoOngEe:
Details will be modified after coordinator - module interface is fixed.

basic_module/staking/src/state.rs
```diff
@@ -0,0 +1,540 @@
...
+    BTreeMap, BTreeSet, HashMap, HashSet,
+};
+use std::ops::Deref;
+
+type KEY = dyn AsRef<[u8]>;
+
+const STAKE_ACCOUNT_PREFIX: [u8; 1] = [0x1];
+const DELEGATION_PREFIX: [u8; 1] = [0x2];
+lazy_static! {
+    static ref METADATA_KEY: Vec<u8> = String::from("Metadata").into_bytes();
```
sgkim126, line 35: Are these should be vectors?
If not, you don't need to use `lazy_static`.
```rust
static METADATA_KEY: &[u8; 8] = b"Metadata";
```

basic_module/staking/src/state.rs
```diff
@@ -0,0 +1,540 @@
...
+impl NextValidators {
+    pub fn load() -> Self {
+        NextValidators(load_with_key(&*NEXT_VALIDATORS_KEY).unwrap_or_default())
+    }
+
+    pub fn save(self) {
+        write_with_key(&*NEXT_VALIDATORS_KEY, self.0)
+    }
+
+    pub fn elect() -> Self {
```
sgkim126, line 282: FYI, currently, @majecty changes the way to store the validators' public keys.
https://github.com/CodeChain-io/foundry/pull/388

basic_module/staking/src/error.rs
```diff
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 pub use crate::runtime_error::Error as RuntimeError;
+pub use crate::syntax_error::Error as SyntaxError;
```
sgkim126, line 18: Would you change the title to `Implement transaction and related types for the staking module`?

basic_module/staking/src/state.rs
```diff
@@ -538,3 +559,14 @@ impl Banned {
         self.0.contains(public)
     }
 }
+
+pub fn get_stakes() -> HashMap<Public, u64> {
+    let stakeholders = Stakeholders::load();
+    let mut result = HashMap::new();
```
sgkim126, line 587:
```rust
    stakeholders.iter().map(|stakeholder| {
        let account = StakeAccount::load(stakeholder);
        let delegation = Delegation::load(stakeholder);
        (*stakeholder, account.balance + delegation.sum())
    }).collect()
```

basic_module/staking/src/execute.rs
```diff
@@ -276,3 +308,105 @@ pub fn change_params(
     metadata.save();
     Ok(outcome)
 }
+
+fn update_validators(validators: NextValidators) -> Result<TransactionExecutionOutcome, Error> {
+    let outcome: TransactionExecutionOutcome = Default::default();
```
sgkim126, line 313: I cannot find a reason to declare this variable here.

basic_module/staking/src/state.rs
HoOngEe: Ah, thank you. I'll apply your suggestion. They don't have to be vectors.

basic_module/staking/src/state.rs
HoOngEe: Thanks, I applied

basic_module/staking/src/error.rs
HoOngEe: Changed

basic_module/staking/src/execute.rs
HoOngEe: The original intention was add events to the outcome as the function proceeds. However, I'm not sure about what events to emit so now empty. I'll move the default() to the return posititon.

basic_module/staking/src/state.rs
HoOngEe: I'll check it out and apply the change.

HoOngEe:
@sgkim126 @majecty 
I checked all the modifications in the pr #388. Before applying the changes in staking module, I have two concerns.
1. `transaction_index` in a "block" could not be resolved from the module level. I naively thought that adding `transaction_index` when executing transactions based on the index in staking-transactions. However, it also produces another complexity. We're now using an interface function `prepare_transactions` which includes transaction executions to determine post-processing transactions(previously implicit) based on the state after execution. After the post-processing transactions are included, the index will be changed so the index when creating a block and the index after being delivered are different. Manually ignoring post-processing transactions when calculating the index is a possible way but unnatural.
2. Host-side validator ordering is important because it matches with the `bitset` in headers. In a modules system a host-side validator only stores a public key and an abstract voting power. To follow the new rule, host should additionally stores `nominated_at_block_number` and `nominated_at_transaction_index`. I don't think it's a general requirements for modules to determine those things as a tie-breaker.

HoOngEe:
After a discussion @majecty 
I realized I misunderstood the boundary the second is no more a problem. The norm of host-side ordering is `public-key` so only staking module is entirely responsible for tie-breaking.
For the first one, I'll rename the variables and apply the change.

PR #336, opened 2020-04-20
Revise module protocols

dynaxis:
This is what @junha1 will implement.

* Removed dependency on crossbeam in favor of simple Receiver trait. Messages are anyway likely queued elsewhere to send to a module, so we don't need to buffer them separately with crossbeam channels.
* Service objects are created at module creation time, and handles of some of them are exchanged at link creation time.

### DIFF
--- a/module/src/link.rs
+++ b/module/src/link.rs
@@ -2,7 +2,7 @@ use std::any::Any;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use crossbeam_channel::{Receiver, Sender};
+use intertrait::CastFrom;
 use linkme::distributed_slice;
 use once_cell::sync;
 use thiserror::Error;
@@ -13,12 +13,12 @@ type Result<T> = std::result::Result<T, Error>;
 ///
 /// [`Linker`]: ./trait.Linker.html
 #[distributed_slice]
-static LINKER_CTORS: [fn() -> Arc<dyn Linker>] = [..];
+pub static LINKERS: [fn() -> Arc<dyn Linker>] = [..];
 
 /// Returns a `Linker` with the given `id`.
 pub fn linker(id: &str) -> Option<Arc<dyn Linker>> {
     static MAP: sync::Lazy<HashMap<&'static str, Arc<dyn Linker>>> = sync::Lazy::new(|| {
-        LINKER_CTORS
+        LINKERS
             .iter()
             .map(|new| {
                 let linker = new();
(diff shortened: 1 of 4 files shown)


module/src/link.rs
```diff
@@ -65,28 +65,49 @@ pub trait Linkable: Send + Sync {
...
-    /// Sets to send a list of handles to the other end on link.
-    ///
-    /// `desc` is encoded in `CBOR`.
-    fn send(&mut self, desc: &[u8]) -> &mut dyn Port;
+pub trait Port: CastFrom {
+    /// Sets to send a list of handles represented by the `ids` to the other end on link
+    /// creation. The `ids` are indices into a list of service objects created when the module
+    /// owning this port is loaded into a sandbox.CBOR map fed
+    /// to the constructor function.
+    fn export(&mut self, ids: &[usize]) -> &mut dyn Port;
```
junha1, line 73: In my implementation of FML, a handle is represented with more info than just `usize`. May I update this later?

junha1, line 73: Also, some module might want to provide an opaque argument `Vec<u8>` for export/importing to provide various ways of interpreting the `ids`. I'm not sure but are you trying to provide such complexity with `load()`?

module/src/link.rs
```diff
@@ -65,28 +65,49 @@ pub trait Linkable: Send + Sync {
...
 /// [`send`]: ./trait.Port.html#tymnethod.send
 /// [`receive`]: ./trait.Port.html#tymnethod.receive
-pub trait Port: Any + 'static {
-    /// Sets to send a list of handles to the other end on link.
-    ///
-    /// `desc` is encoded in `CBOR`.
-    fn send(&mut self, desc: &[u8]) -> &mut dyn Port;
+pub trait Port: CastFrom {
+    /// Sets to send a list of handles represented by the `ids` to the other end on link
+    /// creation. The `ids` are indices into a list of service objects created when the module
```
junha1, line 70: What is difference between a 'handle' and a 'service object'?

module/src/sandbox.rs
```diff
@@ -22,13 +38,22 @@ pub trait Sandboxer {
...
+    /// The `init` serves as configuration parameters for the module-wide initialization,
+    /// and must be CBOR encoded. The `exports` instruct how to instantiate an ordered list
+    /// of service objects to be exported via links. Each item in the `exports` designates
+    /// a call on a module's constructor function, where the first element is name
+    /// of a constructor function, and the second element is arguments to the constructor
+    /// function encoded in CBOR.
     ///
     /// [`Sandbox`]: ./trait.Sandbox.html
-    fn load(&self, hash: &dyn AsRef<Path>) -> Result<Arc<dyn Sandbox>>;
+    fn load(&self, path: &dyn AsRef<Path>, init: &[u8], exports: &[(&str, &[u8])]) -> Result<Arc<dyn Sandbox>>;
```
junha1, line 56: I didn't understand clearly with this. What is the relationship between this `exports` and that in method `Port::export()`?

module/src/link.rs
dynaxis: `usize` is just an index into an array of service objects crated at module creation time. It's not specific to a link implementation.

dynaxis: `export` is just about what will be passed into the other side of the link. Link implementations are free to transfer additional data internally on link creation time.

module/src/link.rs
dynaxis: `Service object` is the actual entity residing in the serving side, while `handle` is what is passed into the other side (remote side).

module/src/sandbox.rs
dynaxis: `exports` are for creating a list of service objects. Once created, all the links may only share some of them. No creation of additional service objects at link creation time.

PR #338, opened 2020-04-21
Remove MemPool benches

Byeongjee:
This bench code is deprecated. Also, since MemPool doesn't do any
computationally expensive job anymore, we don't need a MemPool bench
anymore.


sgkim126:
@Byeongjee Would you apply this change to the master branch?

Byeongjee:
@sgkim126 Okay. I'll cherry-pick it.

PR #346, opened 2020-04-24
[WIP] Informer e2e tests

MSNTCS:


test/src/e2e/informer.test.ts
```diff
@@ -0,0 +1,113 @@
...
+                    JSON.stringify({
+                        jsonrpc: "2.0",
+                        id: 1,
+                        method: "register",
+                        params: []
+                    })
+                );
+            });
+            subscription.once("message", (message: any) => {
+                expect(message).include(JSON.stringify(message1));
```
majecty, line 53: If you use expect in a callback, you cannot check the case when the callback is not called.
For the case, we are using promiseExpect.

How to use it:
1. Declare promiseExpect variable at the start of describe. [see](https://github.com/CodeChain-io/foundry/blob/master/test/src/e2e.long/futureTransaction.test.ts#L23)
2. Wrap promise using promiseExpect.shouldFulfill(). [see](https://github.com/CodeChain-io/foundry/blob/master/test/src/e2e.long/futureTransaction.test.ts#L58)
3. Call promiseExpect.checkFulfilled(); [see](https://github.com/CodeChain-io/foundry/blob/master/test/src/e2e.long/futureTransaction.test.ts#L65)

The checkFulfilled() will throw an error if there are some promises that are not fulfilled.

test/src/helper/spawn.ts
```diff
@@ -240,7 +240,6 @@ export default class CodeChain {
                 `target/${useDebugBuild ? "debug" : "release"}/foundry`,
                 [
                     ...baseArgs,
-                    "--no-informer",
```
majecty, line 243: This line is not related to this commit. Could you move it?

test/src/e2e/informer2.test.ts
```diff
@@ -0,0 +1,77 @@
...
+                    })
+                );
+            });
+            subscription.once("message", (message: Buffer) => {
+                expect(message.includes("result")).to.be.true;
+            });
+            const message2 = {
+                jsonrpc: "2.0",
+                method: "register",
+                params: ":{BlockGeneration"
```
majecty, line 57: Params looks strange. Could you check it?

test/src/helper/spawn.ts
```diff
@@ -211,7 +215,6 @@ export default class CodeChain {
         this._rpc = new RPC(`http://localhost:${this.rpcPort}`, {
             devel: true
         });
-
```
majecty, line 214: This is not related.

test/src/e2e/informer.test.ts
```diff
@@ -77,7 +77,7 @@ describe("Informer", function() {
                     })
                 );
             });
-            subscription.addEventListener("message", (message: any) => {
+            subscription.once("message", (message: any) => {
```
majecty, line 80: This change does not relate to this commit.

test/src/e2e/informer.test.ts
```diff
@@ -94,10 +94,61 @@ describe("Informer", function() {
...
+                        jsonrpc: "2.0",
+                        id: 1,
+                        method: "register",
+                        params: ["PeerAdded"]
+                    })
+                );
+            });
+            let subscribtionId: string;
+            subscription.once("message", (message: string) => {
+                subscribtionId = message.substr(26, 20);
```
majecty, line 118: Could you extract the subscriptionId using `JSON.parse`?
Using `substr` is hard to mange.

test/src/e2e/informer.test.ts
```diff
@@ -94,10 +94,61 @@ describe("Informer", function() {
...
+                expect(message.includes("result")).to.be.true;
+            });
+            expect(
+                await nodeA.rpc.net.connect({
+                    address: address.toString(),
+                    port: nodeB.port
+                })
+            ).to.be.undefined;
```
majecty, line 126: What is the meaning of checking the result with undefined?

test/src/e2e/informer.test.ts
```diff
@@ -94,10 +94,61 @@ describe("Informer", function() {
...
+            while (
+                !(await nodeA.rpc.net.isConnected({
+                    address: address.toString(),
+                    port: nodeB.port
+                }))
+            ) {
+                await wait(500);
+            }
+            subscription.addEventListener("message", (message: Buffer) => {
+                expect(message.includes("PeerAdded")).to.be.true;
```
majecty, line 136: Could you parse the message and extract the "PeerAdded"?

test/src/e2e/informer.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import { expect } from "chai";
+import "mocha";
+import { wait } from "../helper/promise";
+import CodeChain from "../helper/spawn";
+
+describe("Informer", function() {
+    let nodeA: CodeChain;
+    let nodeB: CodeChain;
+    const WebSocket = require("ws");
```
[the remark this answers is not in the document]
majecty: How about adding `informerWs` in the CodeChain type?
It will be greate if we can use websocket `nodeA.informerWS` like `nodeA.rpc`.

test/src/e2e/informer.test.ts
majecty: I recommend you read PromiseExpect code. It is simple.

test/src/helper/spawn.ts
MSNTCS: I moved it to another commit.

MSNTCS:
I tested all 4 categories of tests on my machine. Everything works accordingly. I will rebase my branch after the review.

test/src/e2e/informer.test.ts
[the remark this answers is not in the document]
majecty: @MSNTCS 
The comment below is not applied:
> How about adding informerWs in the CodeChain type?
It will be great if we can use websocket nodeA.informerWS like nodeA.rpc.
Do you have any reason to ignore the comment?

test/src/helper/spawn.ts
majecty: Great!

test/src/e2e/informer.test.ts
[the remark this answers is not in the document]
MSNTCS: Actually, because I did it 2 months ago, I do not remember why. I will try to reflect what you are saying.

test/src/e2e/informer.test.ts
majecty: @MSNTCS Please reply to this comment.

MSNTCS: First of all, I wanted to make sure it is connected and the connection is established. Second, I just follow the same idea that we currently use in [network2.test.ts](https://github.com/CodeChain-io/foundry/blob/49035aee1af5f191ff4581837003667a070a70ee/test/src/e2e/network2.test.ts#L64)

majecty: @MSNTCS Could you check what `to.be.undefiend` is doing?
When the `expect(...).to.be.undefined` fails?
When the `await nodeA.rpc.net.connect(...)` returns undefined ( or other thing)?

test/src/e2e/informer.test.ts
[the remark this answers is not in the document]
MSNTCS: @majecty Do you want me to create a new repository like the foundry-rpc-js repository?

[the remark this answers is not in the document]
majecty: No, that's not my intention.
What I said before is moving WebSocket initialization to `CodeChain` type: https://github.com/CodeChain-io/foundry/blob/49035aee1af5f191ff4581837003667a070a70ee/test/src/helper/spawn.ts#L69

test/src/e2e/informer.test.ts
MSNTCS: @majecty
1- what undefined do is like when the result does not have an actual type.
2- I think in this situation it fails``` expect(1).to.be.undefined```;
3- ```await nodeA.rpc.net.connect(...)``` return void which based on our previous implementation undefined is used to make sure the connection is established.

test/src/e2e/informer.test.ts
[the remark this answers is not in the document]
MSNTCS: Done

test/src/e2e/informer.test.ts
MSNTCS: I think there is no point in what I did while I am checking the connectivity later.

MSNTCS:
@majecty I tested the two failed tests of dynval 1 on my machine and all of them passed.

test/src/e2e/informer.test.ts
majecty: Thanks for researching it.
My opinion is that `expect(...connect...).to.be.undefined` doesn't do anything meaningful.
The `connect` will return undefined or throw an error. Since the function returns only undefined, the expect always passes.
If we want to assert that the `connect` does not throw, we must use `.not.throw`: https://www.chaijs.com/api/bdd/#method_throw
I checked when the `expect(xxx).to.be.undefined` came from. It was an old commit: b60982f28ee81fc50f57744b7b92a4bf7d75b682.

@MSNTCS Could you fix all unnecessary `to.be.undefined` in the later PR?

MSNTCS: @majecty Done

MSNTCS: @majecty Do you want me to change [this PR](https://github.com/CodeChain-io/foundry/commit/b60982f28ee81fc50f57744b7b92a4bf7d75b682)?

majecty: Please create a new PR that removes `to.be.undefined` after this PR is merged.

PR #347, opened 2020-04-26
Remove the concept of account id

sgkim126:
Since the Foundry uses ed25519, it has no merits to use the account id instead of the public key.


sgkim126:
I split the commits into 3 in order to make it easy to review. I'll squash them if you approve.

sgkim126:
@HoOngEe I resolved the conflict.

sgkim126:
I resolved the conflict and merged the commits into one.

Issue #343, opened 2020-04-21
Find an alternative scheme for bech32


sgkim126:
We need a more efficient scheme to encode the address.

sgkim126:
I'll use base64 to encode the key. There are more efficient encoding like base85, but they have a problem that this scheme cannot be used over HTTP natively.

sgkim126:
I changed it to use [base64 URL and filename safe variation](https://tools.ietf.org/html/rfc4648#section-5) without padding.

Issue #342, opened 2020-04-21
Replace the account id to the public key


sgkim126:
Since the Foundry uses ed25519, using it doesn't benefit anymore to use the account id.

PR #348, opened 2020-04-26
Import the source of foundry-primitives-js for e2e tests

sgkim126:
Because of https://github.com/CodeChain-io/foundry/issues/342 and https://github.com/CodeChain-io/foundry/issues/343, I'll change the way to calculate the platform address.
In order to do it, I want to import the source of foundry-primitives-js.
I'll apply changes to https://github.com/CodeChain-io/foundry-primitives-js/ after I finish #342 and #343.


test/src/primitives/test/Hxxx.test.ts
```diff
@@ -1,156 +1,164 @@
...
+        });
 
-    test("ensure", () => {
-        const zero = _.repeat("00", byteLength);
-        expect(Hxxx.ensure(zero)).toEqual(new Hxxx(zero));
-        expect(Hxxx.ensure(new Hxxx(zero))).toEqual(new Hxxx(zero));
-    });
+        it("check", () => {
+            const zero = _.repeat("00", byteLength);
+            expect(Hxxx.check(new Hxxx(zero))).true;
```
majecty, line 38: Wow, I didn't know the usage of `.true`. before.

PR #352, opened 2020-04-28
Resolve a warning file found to be present in multiple build targets

somniumism:
When building mold branch on foundry, we met the warning "file found to be present in multiple build targets". I don't understand the exact cause of this warning, it seems to occur if there is another target in the directory where the target is located, when specifying a build target.

I resolve the warning by creating a new directory `tests/simple` and moving `tests/simple.rs` to `tests/simple/simple.rs`


Byeongjee:
Could you add the reasoning behind this change to the commit message?

somniumism:
@Byeongjee I added. Thanks.

PR #354, opened 2020-04-28
Modify BlockExecutor trait methods' return types

HoOngEe:
Now BlockOutcome contains a reason for a block rejection and
update_validator_set and updated_consensus_params are optional.
Also, execute_transactions returns execution outcome.

### DIFF
--- a/coordinator/src/lib.rs
+++ b/coordinator/src/lib.rs
@@ -48,7 +48,7 @@ impl BlockExecutor for Coordinator {
         unimplemented!()
     }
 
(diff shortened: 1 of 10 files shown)


Byeongjee:
If you change the interfaces, you will have to change the host code as well.
In that case, the commit will not be very small, so I think it's better to separate the commits.

HoOngEe:
I'm almost done but I think a handling for erroneous transactions is needed.

HoOngEe:
I think we'd better to merge this pr first without erroneous transaction handling to apply function signature changes. Maybe the erroneous transaction handling is a part of optimization.

PR #355, opened 2020-04-28
Add header sync test

majecty:
I will enhance this test to check signature verification of precommits later.


majecty:
@sgkim126 Thanks.
I copied the structure fo the onChainHeaderValid.test.ts from onChainBlockValid.test.ts. I found that the hardcoded values are not used. I removed those hard coded values.

majecty:
Since the failed test is "discovery5 nodes", I'll merge this PR.

PR #360, opened 2020-04-29
Upgrade to Rust 1.43.0

### LINKED ISSUE #345: Upgrade to Rust 1.43.0
https://blog.rust-lang.org/2020/04/23/Rust-1.43.0.html

How to: https://github.com/CodeChain-io/foundry/pull/259

HoOngEe:
it resolves #345


core/src/client/test_client.rs
```diff
@@ -438,7 +438,7 @@ impl ImportBlock for TestBlockChainClient {
         }
         let len = self.numbers.read().len();
         if number == len {
-            mem::replace(&mut *self.last_hash.write(), h);
```
sgkim126, line 441: How about using `swap`?

sgkim126: https://doc.rust-lang.org/std/mem/fn.swap.html

HoOngEe: If we use swap here, the following lines are affected from the content swap.
This function returns the value h as `Ok(h)`, `swap` will produce unwanted changes in code.

sgkim126: I think it's okay if we call swap on the cloned `h`, but I'm not sure it's more readable. Let's go ahead with this code.

PR #361, opened 2020-04-30
Change the scheme for the platform address

sgkim126:
The address consists of four parts, checksum, public key, network id, and version.

The version it the last byte of the address.
I limited the version from 0 to 15 to ensure the version can be encoded in a byte with the hex encoding. The current version is 0 and the description below can be changed according to the version.

The network id is 2 bytes.
Currently, it is restricted to use the lowercase alphabets(a-z) only. Since they are printable characters, the scheme doesn't encode it. It's prepended to the version.

The public key is encoded with [base64url](https://tools.ietf.org/html/rfc4648#section-5) without a padding scheme. I choice this encoding because it's efficient and safe to use as a URL or filename. And since we know the length of the public key, we don't need the padding.

The checksum is compression of the public key, network id, and version.
It calculates the 5 bytes with exclusive or operation and uses [the Geohash represntation](https://en.wikipedia.org/wiki/Geohash#Textual_representation) to encode the checksum. I choose it because it is one of the most used no-ambiguity representation.
Since the Geohash representation is 8 bits per 5 bits representation, the length of the checksum is 8 characters.

The below description is the way to calculate the checksum.
```
pubkey := p_0 .... p_31
network_id := n_0 . n_1
c_0 := p_0 . p_1 . p_2 . p_3 . p_4
c_1 := p_5 . p_6 . p_7 . p_8 . p_9
c_2 := p_10 . p_11 . p_12 . p_13 . p_14
c_3 := p_15 . p_16 . p_17 . p_18 . p_19
c_4 := p_20 . p_21 . p_22 . p_23 . p_24
c_5 := p_25 . p_26 . p_27 . p_28 . p_29
c_6 := n_0 . n_1 . version . p_30 . p_31
checksum := c_0 ^ c_1 ^ c_2 ^ c_3 ^ c_4 ^ c_5 ^ c_6
```

If there is a public key looks like `d7a6d266837c1c591383b90d835068b9ed58dd3bcebd6e2859`, the address becomes `01sv1ngs16bSZoN8HFkTg7kNg1Boue1Y3TvOvW4oWRH1jkDOQTwtc0`.
You can read the first eight characters without ambiguity.


majecty:
Could you tell me why the new format uses public key directly?

I agree with including a whole public key in the address. I prefer using public key directly because it is more convenient and it may reduce an additional public key exchange. I fee that the advantage is more important than the security that making a hacker hard to recover a public key from an address.

I want to know your opinion because there can be points that I don't know.

majecty:
I feel funny that the checksum is the name of the address!

HoOngEe:
Here the term "unambiguous" means that geohash excluded "a", "i", "l" and "o" so readers will not be confused when it printed, isn't it?

sgkim126:
@majecty Because ed25519 needs the public key to validate the signature. The RSA can derive the public key from the signature and the message. It's why we used the compressed account id.

@HoOngEe Yes, and it only uses the lower case.

sgkim126:
@kseo I'll merge it and update the spec document if there is no concern.

PR #365, opened 2020-05-06
Move BlockId into codechain-types

Byeongjee:
Coordinator and modules need BlockId, so we should move it to
codechain-types crate in order to avoid cyclic dependecy.


Byeongjee:
@sgkim126 I force pushed it to resolve conflicts. Could you approve it again?

Byeongjee:
@sgkim126 I force pushed to fix conflicts.
Btw, could you teach me a way to fix conflicts without breaking reviews?
If I create a merge commit in my local machine and push it to GitHub, then it automatically removes the merge commit?

PR #366, opened 2020-05-06
Avoid unnecessary nested imports

Byeongjee:
A very minor change just for consistency.


sgkim126:
The travis still fails, but I think it's not caused by this change. I'll merge it.

PR #368, opened 2020-05-06
Remove an unnecessary error type

Byeongjee:
Since we removed unnecessary initialization of genesis header extra
data (c13c5ef899d3062cd718899b07901b4d0dfa72f5), we don't need this
error type anymore.


sgkim126:
It seems that the failed tests are due to timeout. I'll merge it.

PR #371, opened 2020-05-06
Change fetch_transactions_for_block interface

### LINKED ISSUE #370: Change `fetch_transactions_for_block` into `prepare_transactions_with_result`
If we use `TxFIlter::fetch_transactions_for_block`, then proposers have to execute the same transactions twice.
This is because preparing transactions inherently requires the execution of transactions.
We should change the interface in order to avoid the double execution.

Byeongjee:
Resolves #370 
`mem_pool.top_transactions` should be removed.
We may need a separate interface dedicated to handle `ready_transactions` method.

### DIFF
--- a/coordinator/src/lib.rs
+++ b/coordinator/src/lib.rs
@@ -77,7 +77,7 @@ impl TxFilter for Coordinator {
 
     fn filter_transactions<'a>(
         &self,
-        transactions: &'a [&'a TransactionWithMetadata],
+        transactions: Box<dyn Iterator<Item = &'a TransactionWithMetadata> + 'a>,
         memory_limit: Option<usize>,
         size_limit: Option<usize>,
     ) -> (Vec<&'a TransactionWithMetadata>, Vec<&'a TransactionWithMetadata>) {
--- a/coordinator/src/lib.rs
+++ b/coordinator/src/lib.rs
@@ -57,6 +57,14 @@ impl BlockExecutor for Coordinator {
         unimplemented!()
     }
 
+    fn prepare_block<'a>(
+        &self,
+        context: &mut dyn StorageAccess,
+        transactions: Box<dyn Iterator<Item = &'a TransactionWithMetadata> + 'a>,
+    ) -> Vec<&'a Transaction> {
+        unimplemented!()
+    }
+
     fn close_block(&self, context: &mut dyn StorageAccess) -> Result<BlockOutcome, CloseBlockError> {
         unimplemented!()
     }
@@ -64,13 +72,6 @@ impl BlockExecutor for Coordinator {
 
 impl TxFilter for Coordinator {
     fn check_transaction(&self, transaction: &Transaction) -> Result<(), ErrorCode> {
-        unimplemented!()
-    }
-
-    fn fetch_transactions_for_block<'a>(
-        &self,
-        transactions: &'a [&'a TransactionWithMetadata],
-    ) -> Vec<TransactionWithGas<'a>> {
         unimplemented!()
     }
 
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -19,7 +19,7 @@ use crate::error::{BlockError, Error};
 use ccrypto::BLAKE_NULL_RLP;
 use ckey::Ed25519Public as Public;
 use coordinator::traits::BlockExecutor;
-use coordinator::types::{Event, Header as PreHeader, Transaction, VerifiedCrime};
+use coordinator::types::{Event, Header as PreHeader, Transaction, TransactionWithMetadata, VerifiedCrime};
 use cstate::{CurrentValidatorSet, NextValidatorSet, StateDB, StateError, StateWithCache, TopLevelState, TopState};
 use ctypes::header::{Header, Seal};
 use ctypes::util::unexpected::Mismatch;
@@ -28,6 +28,7 @@ use merkle_trie::skewed_merkle_root;
 use primitives::{Bytes, H256};
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
 use std::collections::HashMap;
+use std::iter::Iterator;
 
 /// A block, encoded as it is on the block chain.
 #[derive(Debug, Clone, PartialEq)]
@@ -183,6 +184,16 @@ impl OpenBlock {
         }
         self.block.tx_events = tx_events;
         Ok(())
+    }
+
+    pub fn prepare_block_from_transactions<'a>(
+        &mut self,
+        block_executor: &dyn BlockExecutor,
+        transactions: impl Iterator<Item = &'a TransactionWithMetadata> + 'a,
+    ) {
+        let transactions: Box<dyn Iterator<Item = &'a TransactionWithMetadata>> = Box::new(transactions);
+        let succeeded_transactions = block_executor.prepare_block(self.state_mut(), transactions);
+        self.block.transactions.append(&mut succeeded_transactions.into_iter().cloned().collect());
     }
 
     /// Turn this into a `ClosedBlock`.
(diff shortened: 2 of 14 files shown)


Byeongjee:
The first three commits deal with some minor issues on `mold`.
These changes should have been applied in #260. It was my mistake.

core/src/miner/mem_pool.rs
```diff
@@ -216,33 +216,8 @@ impl MemPool {
...
-            chosen_transactions.iter().map(|tx_with_gas| tx_with_gas.tx_with_metadata.inserted_timestamp).max();
-
-        PendingTransactions {
-            transactions: chosen_transactions
-                .iter()
-                .map(|tx_with_gas| tx_with_gas.tx_with_metadata.tx.clone())
-                .collect(),
-            last_timestamp,
-        }
+    pub fn all_pending_transactions_with_metadata(&self) -> Vec<TransactionWithMetadata> {
```
Byeongjee, line 219: I'm worried about this costly clone.
The best solution would be creating a slice from `transaction_pool.pool`, but unfortunately, it was impossible.
I'm still looking for a nice solution.

Byeongjee: This is why I put **do_not_merge** label.
If you think this is fine, I can just merge it.
Otherwise, I'll continue looking for a solution.

coordinator/src/lib.rs
```diff
@@ -57,6 +57,14 @@ impl BlockExecutor for Coordinator {
         unimplemented!()
     }
 
+    fn prepare_transactions_with_result<'a>(
```
sgkim126, line 60: It seems that it doesn't return the result.

core/src/miner/mem_pool.rs
sgkim126: I think it's okay to merge this commit with leaving the `XXX` comment if fixing it takes too much time.

coordinator/src/lib.rs
Byeongjee: That's right, but as you can see, I actually haven't implemented any of the methods in the file yet.
The implementation would be heavily related to some coordinator-side decisions, so I have to discuss with @dynaxis and other team members.

I think we should implement them soon so I'll have a discussion and work on it, but not in this PR.

Byeongjee:
@sgkim126 I fixed the problem above by just simply returning an `impl Iterator`.
I think returning an iterator is not problematic, and it should be more efficient than the previous revision which returned a cloned vector.

coordinator/src/lib.rs
sgkim126: Even you haven't implemented it yet, you can declare the result type. If you cannot decide how to implement the result type yet, you should not name this method as `prepare_transactions_with_result`.

core/src/block.rs
```diff
@@ -187,6 +187,15 @@ impl<'x> OpenBlock<'x> {
         Ok(())
     }
 
+    pub fn prepare_block_from_transactions(
+        &mut self,
+        block_executor: &dyn BlockExecutor,
+        transactions: &[&TransactionWithMetadata],
```
sgkim126, line 193: How about making this function receives the `Iteration` instead of the slice of the reference??

coordinator/src/lib.rs
Byeongjee: Oh, I misunderstood your concern.
You're worrying because using the term `result` without returning an explicit `Result` type is confusing, right?
Actually I didn't mean this method to return a result type.
The method does two things.
1. Choose transactions to be included in the next block and return them
2. Execute the chosen transactions on the given `context`.

Previously Dohyung suggested the name for a reason in Discord.
Could you suggest a better name based on its behavior?
Maybe `prepare_block` works, because it's easy to infer that this method execute transactions (One of the argument is a mutable reference of `StorageAccess`)

https://discordapp.com/channels/682525519711043609/683931185982079019/699187814977372225
> @byeongjee renaming remove_transactions to filter_transactions seems a good idea. And your idea of splitting methods to open_block, close_block and the others also sounds great. But naming-wise, select_and_execute prefix seems very straight on what it does, but looks a bit verbose to me. And we don’t even select and then execute Txes. Maybe prepare_block would be better, IMHO. But I’m not sure. The exact meaning is ‘select Txes for inclusion in a block to propose by ordering and executing Txes, and then return the root hash’. So if we are to name it to the meaning, it would be more like prepare_transactions_with_result.

core/src/block.rs
Byeongjee: You mean `Iterator`, right?
I'll change the function.
I'm not sure if it's good to use the `Iterator` type in `TxFilter` and `BlockExecutor` methods (because the traits are high-level interfaces exposed to Coordinator) so I didn't change them yet.
Please let me know if you think it's better to apply the change.

coordinator/src/lib.rs
Byeongjee: I used `prepare_block` for now.

core/src/block.rs
sgkim126: Because the `Iterator` is a more abstracted type than the slice, I couldn't understand your concerns.

sgkim126: Did you mean it should be serializable?

core/src/block.rs
```diff
@@ -187,6 +188,16 @@ impl<'x> OpenBlock<'x> {
         Ok(())
     }
 
+    pub fn prepare_block_from_transactions<'a>(
+        &mut self,
+        block_executor: &dyn BlockExecutor,
+        transactions: impl Iterator<Item = &'a TransactionWithMetadata>,
+    ) {
+        let transactions = transactions.collect::<Vec<_>>();
```
sgkim126, line 196: If you change `prepare_block` to receive the iterator also, you don't need to create the vector in here.

core/src/block.rs
Byeongjee: I was slightly worried because the coordinator is directly tied to FML system, and it seemed that it wouldn't be easy to manipulate `Iterator` types in the FML system.

However, as a second thought, if we assume that the coordinator and the host are built and run together in the same process, it wouldn't be problematic to pass `Iterator` type from the host to the coordinator, and it will just be the coordinator's task to unpack the iterator.

In summary, never mind, I'll use iterator type (also in `filter_transactions` method)

core/src/block.rs
Byeongjee: That's true but we still need something complex here, because we have to copy or reference the given iterator somehow because of lifetime parameters of `prepare_block` require it.

sgkim126: You don't need to pass the `Iterator` as `&mut`, `Box<dyn Iterator<Item = &'a TransactionWithMetadata> + 'a>` is enough.
If you think it's complex yet, you can make the `'a` and the Iterator parameters of `BlockExecutor`, like 
```rust
pub trait BlockExecutor<'a, Transactions>: Send + Sync  where Transactions: Iterator<Item = &'a TransactionWithMetadata>
```
But it will make the other code that uses the trait more complex.

Byeongjee: Oh, I didn't think about using a lifetime keyword in the `Box` type. Thanks, I'm still getting used to it.
I actually thought about the latter option, but then I couldn't instantiate `dyn BlockExecutor` object.

Byeongjee: I just realized that I could just use `&mut dyn Iterator` as well.
But I think `Box<dyn Iterator>` is better than `&mut dyn Iterator`, since it states about the reusability more clearly.
I'll use `Box<dyn Iterator>` for both `prepare_block` and `filter_transactions`.

Byeongjee:
@sgkim126 I changed the argument into `Box<dyn Iterator>` type as you advised.
I changed `filter_transactions` method, too, for consistency.

PR #373, opened 2020-05-07
Add trait 'ChainHistoryAccess' to transfer Header from the host side to modules

somniumism:


HoOngEe:
You should omit the module code from the commit

coordinator/src/types.rs
```diff
@@ -269,6 +269,34 @@ pub struct BlockOutcome {
 // Error code returned by check_transaction
 pub type ErrorCode = i64;
 
+/// Uniquely identifies block.
```
HoOngEe, line 272: Let's leave a comment here like
// FIXME: Import the `BlockId` type from the crate "foundry/types"

somniumism: I added the comment, thanks.

HoOngEe: Sorrry, but #365 seems to be merged soon. Let's apply the desirable change in the comment after the pr is merged. It would be merged only after the change.

somniumism: I see. I'll modify it after the pr is merged.

somniumism:
This PR will be pending until #365 is merged and until mold branch is rebased on master branch.

somniumism:
Remaining works:
- [ ] Modify to import the `BlockId` type from the crate `foundry/types`

PR #375, opened 2020-05-07
Implement coordinator-side GraphQL server

Byeongjee:
Coordinator should expose GraphQL handlers of modules through a HTTP connection.


Byeongjee:
@dynaxis For now, I didn't spend much time implementing test cases, because organizing tests took quite more than I expected and I'm not sure about the best practice.
If you suggest me a way to organize them, I'll do it.

In the future, I'll implement my own schema to use the same example in both juniper and async-graphql (and get rid of `starwars` dependency), and add more test cases which cover exception handling logic.

You can still review, merge, and use this PR if you need.
Please review if there is something wrong with http connection, whether it is secure, ...etc.

junha1:
Implemented in `master`.

PR #376, opened 2020-05-07
Fix header RLP encoding in mock code

majecty:
This PR depends on #369

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -32,7 +32,7 @@ use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
 use std::collections::HashSet;
 
 /// A block, encoded as it is on the block chain.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 pub struct Block {
     /// The header of this block
     pub header: Header,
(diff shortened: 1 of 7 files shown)


sgkim126:
LGTM. Let's remove the `do-not-merge` label after #369 merged.

majecty:
@sgkim126 I fixed some test cases.
Since I removed `PartialEq` from `Header`, I couldn't use `rlp_encode_and_decode_test` macro. I used the `Debug` trait instead to compare two structs. Do you think is it okay? It's a weird solution, but I couldn't find a better solution.
https://github.com/CodeChain-io/foundry/pull/376/commits/ffb2ce02039a4a85ab0ae8ca6b79dd518a90aa6e#diff-2253d387d6c10db9b770115468aa0d92R177

```

    /// For a type that does not have PartialEq, uses Debug instead.
    fn assert_eq_by_debug<T: std::fmt::Debug>(a: &T, b: &T) {
        assert_eq!(format!("{:?}", a), format!("{:?}", b));
    }
```

majecty:
The failed test is "connect 100 nodes". I'll merge this PR.

PR #380, opened 2020-05-11
Merge master into mold

Byeongjee:
I requested a review for you because you should know about this change.

For @sgkim126, `open_block_action` and `close_block_action` should be moved to module side, so they are not in this PR.
I also added `update_current_validator_set` and `update_next_block_state` methods in `OpenBlock`. They are equivalent to previous `on_open_block` and `on_close_block`. The host still need implicit state change because it stores parameters and validator sets in the state.

Other changes are quite trivial.


Byeongjee:
@sgkim126 Do you have permission to allow merge commits in non-master branches?
It's currently disabled by the repository setting, even in non-master branches.

majecty:
@Byeongjee @sgkim126 I changed the Foundry repository to allow a merge commit.

PR #382, opened 2020-05-11
Add last_committed_validators field to the Header

HoOngEe:
Modules require information about the committed validators of last
block.

### DIFF
--- a/coordinator/src/types.rs
+++ b/coordinator/src/types.rs
@@ -56,15 +56,24 @@ pub struct Header {
     number: u64,
     /// Block author.
     author: Public,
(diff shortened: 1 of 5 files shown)


HoOngEe:
It can be completed only after the commit `345073` is included in the mold branch.

Byeongjee:
Sorry, I just saw this PR.
I agree with the second commit, but I think OpenBlock should not hold a reference of the engine.
We can just pass engine or last_committed_validators when opening a block.

HoOngEe:
Ok. The other methods take `engine` when it needs its functionality, instead of storing the reference in the struct. I'll make the code consistent with them.

PR #384, opened 2020-05-12
Merge master into mold

somniumism:
Like @Byeongjee did, I requested a review for you because you should know about this merge commit.

#380 was merged, but the master branch still has many changes that are not reflected in the mold branch yet. Therefore, I made the merge commit and also modified to bring `BlockId` from `codechain-types`


test/src/scheme/tendermint-int.json
```diff
@@ -11,37 +11,37 @@
                 "timeoutPrecommitDelta": 1000,
                 "timeoutCommit": 1000,
                 "genesisStakes": {
-                    "tccq8t6d5nxsd7pckgnswusmq6sdzu76kxa808t6m3gtygltrjqeeqncfggwh3": { "stake": 70000,
```
sgkim126: It seems that this change is reverted in the merge commit.

somniumism: I modified it. Could you check it again please?

PR #386, opened 2020-05-13
Verify next_validator_set_hash in header sync

majecty:


majecty:
Since the failed test is "connect 100 nodes", I'll merge this PR.

PR #387, opened 2020-05-13
Do not send validator set if it is same as the previous header's

majecty:


majecty:
@sgkim126 I rebased this PR on the master branch.

PR #388, opened 2020-05-13
Sort validator set in public key order

majecty:
By sorting the validator set using public keys, we can make validator set hash consistent. We can use this property to omit the validator set hash in Header Sync.

If two validators have the same delegation and deposit, the one who became the candidate first will be validator first. We didn't change the rule in this commit. Previously we kept the order of Candidates and Validators to remember who became the candidate first. Now we are saving nomination_starts_at variable to remember it.

### DIFF
--- a/core/src/block.rs
+++ b/core/src/block.rs
@@ -25,7 +25,7 @@ use cstate::{FindDoubleVoteHandler, NextValidators, StateDB, StateError, StateWi
 use ctypes::errors::HistoryError;
 use ctypes::header::{Header, Seal};
 use ctypes::util::unexpected::Mismatch;
-use ctypes::{BlockNumber, TxHash};
+use ctypes::{BlockNumber, TransactionIndex, TxHash};
 use merkle_trie::skewed_merkle_root;
 use primitives::{Bytes, H256};
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
@@ -150,6 +150,7 @@ impl OpenBlock {
         client: &C,
         parent_block_number: BlockNumber,
         parent_block_timestamp: u64,
+        transaction_index: TransactionIndex,
     ) -> Result<(), Error> {
         if self.block.transactions_set.contains(&tx.hash()) {
             return Err(HistoryError::TransactionAlreadyImported.into())
@@ -163,6 +164,7 @@ impl OpenBlock {
             parent_block_number,
             parent_block_timestamp,
             self.block.header.timestamp(),
+            transaction_index,
         ) {
             Ok(()) => {
                 self.block.transactions_set.insert(hash);
@@ -190,8 +192,14 @@ impl OpenBlock {
         parent_block_number: BlockNumber,
         parent_block_timestamp: u64,
     ) -> Result<(), Error> {
-        for tx in transactions {
-            self.push_transaction(tx.clone(), client, parent_block_number, parent_block_timestamp)?;
+        for (index, tx) in transactions.iter().enumerate() {
+            self.push_transaction(
+                tx.clone(),
+                client,
+                parent_block_number,
+                parent_block_timestamp,
+                index as TransactionIndex,
+            )?;
         }
         Ok(())
     }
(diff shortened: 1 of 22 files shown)


core/src/block.rs
```diff
@@ -190,8 +192,17 @@ impl OpenBlock {
...
     ) -> Result<(), Error> {
-        for tx in transactions {
-            self.push_transaction(tx.clone(), client, parent_block_number, parent_block_timestamp)?;
+        for (index, tx) in transactions.iter().enumerate() {
+            self.push_transaction(
+                tx.clone(),
+                client,
+                parent_block_number,
+                parent_block_timestamp,
+                TransactionLocation {
```
sgkim126, line 201: I wonder whether we need to create the transaction location early here.
Only `self_nomiate` needs it and the parent block number is already passed.
How about passing the transaction index here and creating the location when calling the function?

state/src/impls/top_level.rs
```diff
@@ -314,6 +316,7 @@ impl TopLevelState {
         parent_block_number: BlockNumber,
         parent_block_timestamp: u64,
         current_block_timestamp: u64,
+        transaction_location: Option<TransactionLocation>,
```
sgkim126, line 319: Why is the location optional?

sgkim126: Okay. I see the e2e tests.
Some e2e tests postulate the scenario that transactions in the same block. So it can make testing what you really want to do impossible to make all transaction index in the e2e test zero.
But, fixing all e2e tests can be a time-consuming job. How about leaving issue and TODO comment here and making another PR for it?

core/src/block.rs
majecty: I want to distinguish TransactionLocation and parent_block_number.
In the RPC method "chain_executeTransaction" does not have TransactionLocation but can have parent_block_number.

However, I found that Foundry doesn't have **chain_executeTransaction** (Foundry had it before) in Foundry. If we can ensure that all transactions are executed in a block, and always has an index, I agree with you.

Do you know whether the transaction always has transaction index or not?

state/src/impls/top_level.rs
majecty: I assumed that chain_executeTransaction RPC will use the function with `None`. If we can ensure that all transactions have an index, I don't want to use the Option.

Similar:
https://github.com/CodeChain-io/foundry/pull/388/files#r426344706

core/src/block.rs
sgkim126: We removed `chain_executeTransaction`, but it will re-implemented with the module system.
So I cannot say that all transaction is executed in the block, but I think it can be bypassed by adding the transaction index as an argument of the RPC.

sgkim126: or how about to fix the transaction index as zero when the transaction is executed by the RPC?

majecty: @sgkim126 
Passing index or using zero looks reasonable.
I'll fix this PR like followings:
* Use TransactionLocation only when there is no parent block number.
* Use TransactionIndex instead of Option<TransactionLocation>

state/src/impls/top_level.rs
sgkim126: How about adding the index as an argument of the RPC or fixing the transaction index as a constant value(I think 0 is reasonable) if the transaction is executed by the RPC.

sgkim126: Or how about making the argument of the RPC optional and converting it to zero if the argument is null?

majecty: Passing a index or using 0 as a default value looks good to me.
I'll change this PR like this: https://github.com/CodeChain-io/foundry/pull/388#discussion_r426347279

majecty:
@sgkim126 I fixed them.

majecty:
Failed tests succeed in my local MacOS and Linux machines except discovery 5 test.

core/src/consensus/validator_set/dynamic_validator.rs
```diff
@@ -24,23 +24,36 @@ use ctypes::transaction::Validator;
 use ctypes::util::unexpected::OutOfBounds;
 use ctypes::BlockHash;
 use parking_lot::RwLock;
+use std::cmp::Reverse;
```
sgkim126, line 27: :+1:

majecty:
I merged the commits that change the test into the commit that changes the implementation!

majecty:
It seems that this PR makes tests to use more time. I'll dig into it.

PR #391, opened 2020-05-15
[WIP] Fix some minor problems in Account module

Byeongjee:
I'll finish it after merging https://github.com/CodeChain-io/foundry/pull/390
This includes
- Fix return of `execute_transactions` (It should not return an empty vector)
- Rename & organize error types
- Perform more tests in `check_transactions`
- Rename `address: &Public` into `account_id: &AccountId`
- Replace `Vec<u8>` with `primitives::Bytes`

### DIFF
--- a/basic_module/account/src/internal.rs
+++ b/basic_module/account/src/internal.rs
@@ -29,7 +29,7 @@ pub fn add_balance(address: &Public, val: u64) {
     let mut account: Account = get_account(address);
 
     account.balance += val;
-    context.set(address, account.to_vec());
+    context.set(address, account.to_bytes());
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


basic_module/account/src/internal.rs
```diff
@@ -38,11 +38,14 @@ pub fn sub_balance(address: &Public, val: u64) -> Result<(), Error> {
...
-        return Err(Error::InvalidValue(account.balance, val))
+        return Err(Error::InsufficentBalance {
+            balance: account.balance,
+            withdrawal: val,
+        })
     }
 
     account.balance -= val;
-    context.set(address, account.to_vec());
+    context.set(address, account.to_bytes());
```
junha1, line 48: Btw, why not `serde_something::to_vec(&account)` here?

Byeongjee:
We don't need this PR anymore since the Account module has changed a lot.

PR #393, opened 2020-05-16
Use parking_lot instead of the standard sync primitives

sgkim126:
We use `parking_lot` instead of the standard Mutex.
It's more efficient and easy to use.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -355,6 +355,7 @@ dependencies = [
  "ipc-channel",
  "nix",
  "once_cell",
+ "parking_lot 0.6.4",
  "primitives",
  "serde",
  "serde_cbor",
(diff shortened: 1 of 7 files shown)


somniumism:
It has nothing to do with this PR, but the Distribution module use `std::sync::RwLock`. Do we have to change this to `parking_lot::RwLock`? I checked that `parking_lot` provides `RwLock` too.

sgkim126:
@somniumism Thank you. I had missed it. Please let me know if there are other things that use the standard synchronization primitives.
FYI, I know `util/io` and the verification queue are using the standard. They will be fixed someday.

somniumism:
Thank you too. I checked the current prototype of the Distribution module uses `std::sync::Mutex` and `std::sync::RwLock`. I'll modify it.

PR #398, opened 2020-05-20
Fix few things in test code

majecty:
@sgkim126 Could you check "Make Foundry makes blocks slowly if RUN_ON_TEST is set" commit carefully. I'm not sure using an environment variable and the `sleep` method is good or not.


core/src/miner/miner.rs
```diff
@@ -516,6 +516,14 @@ impl MinerService for Miner {
             }
         };
 
+        if std::env::var("RUN_ON_TEST").is_ok() {
```
sgkim126, line 519: :+1:

PR #405, opened 2020-05-24
Importing the staking module

sgkim126:


basic_module/staking/src/impls.rs
```diff
@@ -0,0 +1,123 @@
...
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+struct ABCIHandle {
+    executing_block_header: RefCell<Header>,
+}
+
+impl AdditionalTxCreator for ABCIHandle {
+    fn create(&self) -> Vec<Transaction> {
+        let mut transactions = create_open_block_transactions();
```
sgkim126, line 38: The original code in the mold branch is as below.
```rust
create_open_block_transactions()
    .into_iter()
    .chain(create_close_block_transactions(&*self.executing_block_header.borrow()).into_iter())
    .collect()
```
I changed it because the code above is inefficient.

basic_module/staking/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+fn deserialize<T: serde::de::DeserializeOwned>(buffer: Vec<u8>) -> T {
+    serde_cbor::from_slice(&buffer).unwrap()
+}
+
+fn serialize<T: serde::ser::Serialize>(data: T) -> Vec<u8> {
+    serde_cbor::to_vec(&data).unwrap()
+}
+
+lazy_static! {
+    static ref NETWORK_ID: Mutex<Option<NetworkId>> = Default::default();
```
sgkim126, line 51:
```rust
static ref NETWORK_ID: types::NetworkId = Default::default();
```
`NETWORK_ID` is declared as the above in the mold branch but I changed it because there is a FIXME comment `network_Id should be mutable`, and change `check_network_id` to set `NETWORK_ID` at the first time.

majecty:
Since the failed tests are "discovery 5 nodes" and "connect 100 nodes", I'll merge this PR.

PR #418, opened 2020-06-08
Fetch consistent state in mempool functions

majecty:
We found that mempool mis-calculates sequence in the TPS test. After
fixing the mempool to read consistent state, mempool calculates
sequence well.

Copied from https://github.com/CodeChain-io/codechain/pull/1976


majecty:
Since the failed tests are "bootstrap 100 nodes" and "discovery5", I'll merge this PR.

PR #421, opened 2020-06-21
Change the mutability of SubStorageAccess

sgkim126:
I found the SubStorageAccess takes the immutable references. But, it's safer to take the mutable reference when it changes the internal value because it makes only one thread be able to mutate the values.

### DIFF
--- a/coordinator/src/context/sub_storage_access.rs
+++ b/coordinator/src/context/sub_storage_access.rs
@@ -20,14 +20,14 @@ pub type Value = Vec<u8>;
 // Interface between each module and the coordinator
 pub trait SubStorageAccess {
     fn get(&self, key: &Key) -> Option<Value>;
-    fn set(&self, key: &Key, value: Value);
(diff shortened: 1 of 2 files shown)


coordinator/src/context/sub_storage_access.rs
```diff
@@ -20,14 +20,14 @@ pub type Value = Vec<u8>;
...
     /// Create a recoverable checkpoint of this state
-    fn create_checkpoint(&self);
+    fn create_checkpoint(&mut self);
     /// Revert to the last checkpoint and discard it
-    fn revert_to_the_checkpoint(&self);
+    fn revert_to_the_checkpoint(&mut self);
     /// Merge last checkpoint with the previous
-    fn discard_checkpoint(&self);
+    fn discard_checkpoint(&mut self);
 }
```
majecty, line 33: The same issue as https://github.com/CodeChain-io/foundry/pull/422#discussion_r443278927

PR #422, opened 2020-06-21
Clean-up the account module

sgkim126:


basic_module/account/src/core.rs
```diff
@@ -25,17 +25,20 @@ pub trait CheckTxHandler {
...
 
-    fn sub_balance(&self, account_id: &Public, val: u64) -> Result<(), Error>;
+    fn sub_balance(&mut self, account_id: &Public, val: u64) -> Result<(), Error>;
 
-    fn set_balance(&self, account_id: &Public, val: u64);
+    fn set_balance(&mut self, account_id: &Public, val: u64);
 
-    fn increment_sequence(&self, account_id: &Public);
+    fn increment_sequence(&mut self, account_id: &Public);
 }
```
majecty, line 42: If the traits "TransactionExecutor" and "AccountManager" are used in the module boundray, it can't use &mut self in the final version. `mut` keyword is possible only when modules are implemented in Rust and included in the same binary.

We may use the mut for now and remove it later. What is your opinion?

sgkim126: Sorry, I didn't understand your concern.
The `mut` keyword is a way to guarantee memory safety in Rust, so 
if the module is implemented in another language, of course, it should find a way to guarantee it. And the implementation in another language can't import a trait; the communication between the modules is defined by the protocol that uses IPC(or socket or something that can be serialized), not by traits. I think the trait should define in the way of Rust idiom.

majecty: Yes normally trait is used in Rust program and we use separate IPC protocol.
However, we are implementing "Remote Trait Object" as an IPC protocol. The "Remote Trait Object" uses a "trait" as a protocol interface. The "Remote Trait Object" generates serialize, deserialize, and remote procedure call code by reading the "trait" code. And it can't handle the "mut" semantic.

Using "mut" for now looks okay. I wanted to say that it will be removed when we introduce "Remote Trait Object" in the module communication.

junha1: In `remote-trait-object` context, trait method's `&self` represents immutability of **Service object itself**, not the semantically underlying object. Possible implementor of `trait AccountManager` would be even stateless; it doesn't even have anything to mutate.

sgkim126: I don't know how do you implement the AccountManager, but you cannot make it stateless. The accounts are not global values, the states of the accounts depend on the current state root. So the context should not be global and each context should mark which state root it currently looks up. Because the context keeps the current state, it is dangerous to make set and remove take a mutable self.

junha1: @sgkim126 I'm talking about a state (not the blockchain state) of the **service object itself**. Sorry for the confusing terminology.

majecty: Actually, I don't know `AccountManager` will be used as remote-trait-object or not.
We can consider it when we introduce remote-trait-object.
So, I'm ok with the current PR status.

@sgkim126 @junha1 Please discuss without me.

PR #426, opened 2020-06-26
Modify connection assertion in network tests

MSNTCS:


test/src/e2e/network2.test.ts
```diff
@@ -61,7 +61,7 @@ describe("network2 nodes", function() {
                     address: address.toString(),
                     port: nodeB.port
                 })
-            ).to.be.undefined;
+            ).not.throw;
```
majecty, line 64: Great!

PR #427, opened 2020-06-26
Add missing sequence handling

junha1:


timestamp/src/account.rs
```diff
@@ -58,20 +58,46 @@ impl AccountManager for Context {
...
+                if default {
+                    self.create_account(public).expect("Synchronization bug in AccountManager");
+                    self.increase_sequence(public, false).expect("Synchronization bug in AccountManager");
+                    Ok(())
+                } else {
+                    result
+                }
+            }
+            _ => result,
+        }
```
majecty, line 100: How about this?
```suggestion    fn increase_sequence(&self, public: &Public, default: bool) -> Result<(), Error> {
        let option_bytes = self.storage.read().get(public);
        if option_bytes.is_none() {
            if default {
                self.create_account(public).expect("Synchronization bug in AccountManager");
                self.increase_sequence(public, false).expect("Synchronization bug in AccountManager");
                return Ok(())
            } else {
                return Err(Error::NoSuchAccount)
            }
        }

        let bytes = option_bytes.unwrap();
        let mut account: Account = serde_cbor::from_slice(&bytes).map_err(|_| Error::InvalidKey)?;
        account.seq += 1;
        self.storage.read().set(public, serde_cbor::to_vec(&account).unwrap());
        Ok(())
    }
```

junha1:
@majecty I applied your suggestions

PR #428, opened 2020-06-30
Add tests and token list up

junha1:


timestamp/src/token.rs
```diff
@@ -127,6 +149,11 @@ impl Context {
         self.storage.read().set(&Self::get_key(key), serde_cbor::to_vec(account).unwrap());
     }
 
+    /// set_owning_accounts_with_issuer() must not fail
+    fn set_owning_accounts_with_issuer(&self, issuer: H256, set: HashSet<Public>) {
+        self.storage.read().set(&Self::get_key_account_set(issuer), serde_cbor::to_vec(&set).unwrap());
```
majecty, line 154: How about using BTreeSet instead of HashSet?

If two hash sets have the same elements, do they return elements with **the same order**?

Normally hash set's property is not keeping its order. Since I didn't check the detail of Rust HashSet, it may return consistent results of elements order. IMO, using a data structure that has a property that ensures the elements' order is required.

junha1: Thanks. I replaced it with `BTreeSet`

PR #429, opened 2020-07-01
Implement staking module

junha1:


timestamp/src/staking.rs
```diff
@@ -0,0 +1,97 @@
...
+            .collect()
+    }
+}
+
+impl InitGenesis for Context {
+    // _storage will be removed in the updated interface
+    fn init_genesis(&self, config: &[u8], _storage: Box<dyn SubStorageAccess>) {
+        let initial_validator_set: Validators = serde_cbor::from_slice(config).unwrap();
+        for public in initial_validator_set {
+            self.token_manager.read().issue_token(&self.validator_token_issuer, &public).unwrap();
```
majecty, line 75: It's not important for now. We may change this later.
Isn't it `issue_token`be mutable?

junha1: @majecty Yes, but it was designed to be `&self` because RTO didn't support mutablity back then.

junha1: We have to revisit mutablity of all methods someday

PR #430, opened 2020-07-06
Implement process sandbox

junha1:


module/src/sandbox.rs
```diff
@@ -85,6 +85,8 @@ pub trait Sandboxer: Send + Sync {
 pub trait Sandbox: Linkable {
     /// Returns the `Sandboxer` for this sandbox.
     fn sandboxer(&self) -> Arc<dyn Sandboxer>;
+
+    fn debug(&mut self, arg: &[u8]) -> Vec<u8>;
```
majecty, line 89: It seems that `debug` is not a mandatory function for the `Sandbox`.
How about giving a default implementation?

module/src/impls/process.rs
```diff
@@ -0,0 +1,250 @@
...
+    type Execution = fproc_sndbx::execution::executor::PlainThread;
+    fn is_intra() -> bool {
+        true
+    }
+}
+
+#[remote_trait_object_macro::service]
+pub trait SandboxForModule: remote_trait_object::Service {
+    fn ping(&self);
+}
```
majecty, line 105: Do you have any plan to add functions in the `SandboxForModule`?
If not, we don't have to use rto context between sandboxee and sandboxer.

module/src/sandbox.rs
junha1: Done

PR #452, opened 2020-07-31
Implement the voting system 

MSNTCS:


votingsystem/src/common.rs
```diff
@@ -0,0 +1,71 @@
...
+impl fmt::Display for NetworkId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        let s = std::str::from_utf8(&self.0).expect("network_id a valid utf8 string");
+        write!(f, "{}", s)
+    }
+}
+
+impl Default for NetworkId {
+    fn default() -> Self {
+        NetworkId([116, 99])
```
majecty, line 35: Let's use character literal.
```suggestion
        NetworkId([b't', b'c'])
```

I chanted the same code in timestamp app: https://github.com/CodeChain-io/foundry/pull/453/files

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn execute_transaction(&mut self, transaction: &Transaction) -> Result<TransactionOutcome, ()> {
+        if let Err(error) = self.excute_tx(transaction) {
+            match error {
+                ExecuteError::InvalidMetadata => Err(()),
+                ExecuteError::InvalidSign => Err(()),
+                ExecuteError::InvalidFormat => Err(()),
+            }
+        } else {
+            Ok(Default::default())
+        }
```
majecty, line 117: Why didn't you call `execute_tx` in this function?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    fn block_closed(&mut self) -> Result<Vec<Event>, CloseBlockError> {
+        Ok(Vec::new())
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct TimeStamp{
+    time: u64,
+}
```
majecty, line 136: It will be serialized as `{ time: 3 }` instead of `3`.
Is it your intention?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+impl TimeStamp {
+    fn set_time(&mut self, time: u64){
+        self.time = time;
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct GeneralMeetingId {
+    id: u8,
+}
```
majecty, line 148: It will be serialized as `{ id: 3 }` instead of `3`.
Is it your intention?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub enum GeneralMeetingError {
+    NoSuchMeeting,
+    MeetingExist,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct GeneralMeeting {
+    id: GeneralMeetingId,
+    number_of_options: u32,
```
majecty, line 177: Do we need `number_of_options`?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum GeneralMeetingError {
+    NoSuchMeeting,
+    MeetingExist,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct GeneralMeeting {
```
majecty, line 175: Where is `number_of_agendas`?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+    voting_end_time: TimeStamp,
+    tallying_time: TimeStamp,
+    result: Option<Vec<VoteResult>>,
+}
+
+impl GeneralMeeting {
+    pub fn new(end_time: TimeStamp, tallying_time: TimeStamp) -> Self {
+        let meeting_id= GeneralMeetingId::new();
+        let options = 4;
+        let final_result = Some(Vec::new());
```
majecty, line 187: If there is no final_result, isn't using `None` better?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+impl InitGenesis for Context {
+    fn begin_genesis(&mut self) {}
+
+    fn init_genesis(&mut self, config: &[u8]) {
+    }
+
+    fn end_genesis(&mut self) {}
+}
```
majecty, line 100: If you don't do anything in functions in InitGenesis, why do you implement it?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    fn check_transaction(&self, transaction: &Transaction) -> Result<(), coordinator::types::ErrorCode> {
+        let todo_fixthis: coordinator::types::ErrorCode = 3;
+        assert_eq!(transaction.tx_type(), "CreateGeneralMeeting");
+        let tx: OwnTransaction = serde_cbor::from_slice(&transaction.body()).map_err(|_| todo_fixthis)?;
+        tx.verify().map_err(|_| todo_fixthis)?;
+        Ok(())
+    }
```
majecty, line 126: Shouldn't we check the signer?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+
+enum ExecuteError {
+    InvalidMetadata,
+    InvalidSign,
+    InvalidFormat,
+}
+
+struct Context{
+    pub storage: Option<Box<dyn SubStorageAccess>>,
+    pub meeting: Option<Box<dyn MeetingManager>>,
```
majecty, line 39: Why did you add `meeting` in Context?
It is not a remote service.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,262 @@
...
+            tallying_time,
+            result: final_result
+        }
+    }
+}
+
+#[service]
+pub trait MeetingManager: Service {
+    fn create_general_meeting(&mut self, general_meeting: GeneralMeeting, id: GeneralMeetingId) -> Result<GeneralMeetingId,GeneralMeetingError>;
+}
```
majecty, line 201: Why did you create `MeetingManager` service?
Which module will use this service?

votingsystem/src/general_meeting.rs
MSNTCS: Here is the part that is not complete and needs our discussion because I do not have a good insight of InitGenesis.

votingsystem/src/general_meeting.rs
MSNTCS: A verified person must be allowed to create a general meeting.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,271 @@
...
+            id: random_id
+        }
+    }
+}
+
+impl AsRef<[u8]> for GeneralMeetingId{
+    fn as_ref(&self)->&[u8]{
+        self.as_ref()
+    }
+}
```
majecty, line 171: Did you add `as_ref` for the serialization? I think AsRef is not related to the serialization.
You can use `#[serde(transparent)]` to serialize a struct that has only a field exactly the same as if its one field were serialized itself.

https://serde.rs/container-attrs.html#transparent

I recommend you to read the serde document. It isn't long.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,271 @@
...
+enum ExecuteError {
+    InvalidMetadata,
+    InvalidSign,
+    InvalidFormat,
+    InvalidAdmin,
+}
+
+struct Context{
+    pub storage: Option<Box<dyn SubStorageAccess>>,
+    pub admin: Option<Public>,
```
majecty, line 42: How about reading `admin` in the BlockChain state always?
If the admin value in the state changes, the GeneralMeeting module must read the value.
Currently, we are not changing it, but I predict it will be changed.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,271 @@
...
+    result: Option<Vec<VoteResult>>,
+}
+
+impl GeneralMeeting {
+    pub fn new(end_time: TimeStamp, tallying_time: TimeStamp) -> Self {
+        let meeting_id= GeneralMeetingId::new();
+        let options = 4;
+        Self{
+            id: meeting_id,
+            number_of_agends: options,
```
majecty, line 213: I used the term "agenda" as the number of votes in the meeting.
For example, if there are two agendas in the meeting like "Do you like Kimchi" and "Do you know Bibimbap", the voters should vote two times like "I don't like kimchi" and "I don't know Bibimbap".

So the number of agenda cannot be hard-coded. Someone who creates a general meeting could set the number of agendas in the meeting.

votingsystem/src/general_meeting.rs
majecty: OK, I understand your intention.

I thought you used as_ref for the value serialization.
How about naming the [u8] as `Key` to express your intention?

```rust
type StorageKeyRef<'a> = &'a [u8];

impl AsRef<StorageKey> for GeneralMeetingId{
    fn as_ref(&self)->StorageKeyRef {
      std::slice::from_ref(&self.id)
    }
}
```

Also calling `self.as_ref()` in the as_ref definition is a infinite loop. If you want to make &T to &[T] you should use `std::slice::from_ref`: https://doc.rust-lang.org/std/slice/fn.from_ref.html

votingsystem/src/general_meeting.rs
majecty: @MSNTCS Could you check this comment?

votingsystem/src/general_meeting.rs
majecty: @MSNTCS Could you check this comment?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,248 @@
...
+        self.storage.replace(storage.unwrap_import().into_remote());
+    }
+}
+
+impl InitGenesis for Context {
+    fn begin_genesis(&mut self) {}
+
+    fn init_genesis(&mut self, config: &[u8]) {
+        let admin: Public = serde_cbor::from_slice(&config).unwrap();
+        self.storage_mut().set(admin.as_ref(), serde_cbor::to_vec(&admin).unwrap());
```
majecty, line 95: Why did you use the admin's public key as the "key"?
How will use read the value?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    fn check_transaction(&self, transaction: &Transaction) -> Result<(), coordinator::types::ErrorCode> {
+        let todo_fixthis: coordinator::types::ErrorCode = 3;
+        assert_eq!(transaction.tx_type(), "GeneralMeeting");
+        let tx: OwnTransaction = serde_cbor::from_slice(&transaction.body()).map_err(|_| todo_fixthis)?;
+        tx.verify().map_err(|_| todo_fixthis)?;
+
+        let admin_public_key = tx.signer_public;
```
majecty, line 125: Shouldn't we check the admin_public key is really admin?

votingsystem/src/general_meeting.rs
MSNTCS: Yes, that is what I meant.

votingsystem/src/general_meeting.rs
MSNTCS: Yes that is my intention.

votingsystem/src/general_meeting.rs
majecty: If we serialize a GeneralMeeting struct, it will be serialized like this:

```json
{
  tallying_time: {
    time: 3
  }
}
```

I think the format below is better.

```json
{
  tallying_time: 3
}
```

What do you think about it?

Also we can do it by adding a line over `Timestamp` struct: `#[serde(transparent)] `
https://serde.rs/container-attrs.html#transparent

MSNTCS: I agree with your point. I will use ``#[serde(transparent)]``. Thanks for mentioning.

votingsystem/src/general_meeting.rs
MSNTCS: @majecty, please read the next three lines.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,251 @@
...
+        self.storage.replace(storage.unwrap_import().into_remote());
+    }
+}
+
+impl InitGenesis for Context {
+    fn begin_genesis(&mut self) {}
+
+    fn init_genesis(&mut self, config: &[u8]) {
+        let admin: Public = serde_cbor::from_slice(&config).unwrap();
+        self.storage_mut().set("admin".as_bytes(), serde_cbor::to_vec(&admin).unwrap());
```
majecty, line 95: Let's declare a constant variable for "admin"

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+    InvalidFormat,
+    InvalidAdmin,
+}
+
+pub enum Id {
+    VotePaperId(u8),
+    BallotId(u8),
+}
+#[derive(Debug, Serialize, Deserialize)]
+pub struct VoteId {
```
majecty, line 44: Let's use serde::transparent.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+    InvalidMetadata,
+    InvalidSign,
+    InvalidFormat,
+    InvalidAdmin,
+}
+
+pub enum Id {
+    VotePaperId(u8),
+    BallotId(u8),
+}
```
majecty, line 42: Where will you use this `Id`?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,251 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct TimeStamp {
+    time: u64,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(transparent)]
+pub struct GeneralMeetingId {
+    pub id: u8,
```
majecty, line 148: u8 is too small. How about U256?

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct VoteId {
+    id: u8,
```
majecty, line 45: u8 is too small.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl AsRef<StorageKeyRef> for VoteId {
+    fn as_ref(&self) -> &[u8] {
+        std::slice::from_ref(&self.id)
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct BallotId {
+    id: u8,
```
majecty, line 58: u8 is too small.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+        std::slice::from_ref(&self.id)
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct VotePaper {
+    general_meeting_id: GeneralMeetingId,
+    agenda_number: u32,
+    authorization_token: u64,
+    used_in: VoteId,
```
majecty, line 71: Please use `Option<VoteId>`. There will be vote papers that are not used yet.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+        if transaction.tx_type() != "Vote" {
+            return Err(ExecuteError::InvalidMetadata)
+        }
+        let tx: OwnTransaction =
+            serde_cbor::from_slice(&transaction.body()).map_err(|_| ExecuteError::InvalidFormat)?;
+        tx.verify().map_err(|_| ExecuteError::InvalidSign)?;
+
+        let mut meeting_id = &GeneralMeetingId {
+            id: 0,
+        };
```
majecty, line 163: It looks strange.
The meeting_id is set in the "CreateVotePaper" action and used in the "Vote" action.
The can't be executed at the same function.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+        match tx.tx.action {
+            TxAction::CreateVotePaper(TxCreateVotePaper {
+                general_meeting_id,
+                agenda_number,
+                authorization_token,
+                voter_public_key,
+            }) => {
+                let vote_paper =
+                    VotePaper::new(general_meeting_id.clone(), agenda_number, authorization_token, voter_public_key);
+                self.storage_mut().set(vote_paper.used_in.as_ref(), serde_cbor::to_vec(&vote_paper).unwrap());
```
majecty, line 173: When we create a vote paper, the vote_paper_used_in will be empty because it is not used yet.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,270 @@
...
+                meeting_id = &general_meeting_id;
+                Ok(Id::VotePaperId(vote_paper.used_in.id))
+            }
+            TxAction::Vote(Ballot {
+                choice,
+                voter_signature,
+            }) => {
+                let ballot = Vote::new(choice, voter_signature);
+                let id = &ballot.vote_paper_id;
+                self.storage_mut().set(id.as_ref(), serde_cbor::to_vec(&ballot).unwrap());
```
majecty, line 183: This will remove the vote paper information. Why did you remove the vote paper form the state?

votingsystem/src/voting.rs
MSNTCS: Done

votingsystem/src/general_meeting.rs
MSNTCS: Cannot make the admin constant due to the fact that the config value is not constant.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,280 @@
...
+            Ok(Default::default())
+        }
+    }
+
+    fn check_transaction(&self, transaction: &Transaction) -> Result<(), coordinator::types::ErrorCode> {
+        let todo_fixthis: coordinator::types::ErrorCode = 3;
+        assert_eq!(transaction.tx_type(), "Vote");
+        let tx: OwnTransaction = serde_cbor::from_slice(&transaction.body()).map_err(|_| todo_fixthis)?;
+        tx.verify().map_err(|_| todo_fixthis)?;
+        Ok(())
```
majecty, line 230: Please check these things.

TxAction::CreateVotePaper:

* general meeting is exist or not
* general meeting is finished or not
* voting time is finished or not
* agenda number exceeds the agenda count in the general meeting or not

TxAction::Vote

* vote paper is already used or not
* choice should be one of 4 choices
* voter signature is generated from voter public key or not

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+impl Service for Context {}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct TxCreateVotePaper {
+    general_meeting_id: GeneralMeetingId,
+    agenda_number: u32,
+    authorization_token: H256,
```
majecty, line 138: What is the `authorization_token`?

MSNTCS: It is a token that the person gets after authentication. We should check whether the person is authorized or not but because we do not save their tokens in the first place it needs a mechanism for it.

votingsystem/src/common.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+impl Default for NetworkId {
+    fn default() -> Self {
+        NetworkId([b't', b'c'])
+    }
+}
+
+pub trait Action: Serialize + std::fmt::Debug {}
```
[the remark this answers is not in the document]
MSNTCS: @majecty if you remove `Serialization` here and replace it with `Desrilization`, you would face a `life-time` error.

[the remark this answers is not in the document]
majecty: Thanks, I'll check it.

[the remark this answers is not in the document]
majecty: Sorry, serialzation is required to get the hash of the transaction.
https://github.com/CodeChain-io/foundry/pull/452/files#diff-725c366433452466115179ac7d399f0fR67 Let's keep it.

votingsystem/src/voting.rs
majecty: I think my explanation was not good.

We can think about it like this:
Actually, creating a vote paper (with the public key) means that a "person" who has a signature is a stakeholder.
When a voter sends vote transaction, we can verify the vote by checking the signature is created by the "private key".

votingsystem/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pub const prefix: &str = "VoteBox";
+
+pub mod common;
+pub mod general_meeting;
+pub mod voting;
+
+pub fn generate_key(meeting_id: &general_meeting::GeneralMeetingId) -> Vec<u8> {
```
majecty, line 23: `generate_key` generates a key for "VoteBox", but the name does not be related to the "VoteBox".
Please make the name contains the "VoteBox" information.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,382 @@
...
+                        vote_result.favor += number_of_shares;
+                    }
+                    if vote_choice == crate::voting::against {
+                        vote_result.against += number_of_shares;
+                    }
+                    if vote_choice == crate::voting::absention {
+                        vote_result.absention += number_of_shares;
+                    }
+                }
+                Ok(TxResult::Result(vote_result))
```
majecty, line 140: Shouldn't we save the vote_result in `GeneralMeeting` struct?

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,382 @@
...
+                ExecuteError::InvalidAdmin => Err(()),
+            }
+        } else {
+            Ok(Default::default())
+        }
+    }
+
+    fn check_transaction(&self, transaction: &Transaction) -> Result<(), coordinator::types::ErrorCode> {
+        let todo_fixthis: coordinator::types::ErrorCode = 3;
+        assert_eq!(transaction.tx_type(), "GeneralMeeting");
```
majecty, line 183: Let's make a constant for "GeneralMeeting"

votingsystem/src/general_meeting.rs
majecty, line 140: Please check the block's timestamp is later than the `tallying_time`.

votingsystem/src/general_meeting.rs
```diff
@@ -0,0 +1,382 @@
...
+                    let number_of_shares = self.vote().get_shares(&vote_id).unwrap();
+                    if vote_choice == crate::voting::favor {
+                        vote_result.favor += number_of_shares;
+                    }
+                    if vote_choice == crate::voting::against {
+                        vote_result.against += number_of_shares;
+                    }
+                    if vote_choice == crate::voting::absention {
+                        vote_result.absention += number_of_shares;
+                    }
```
majecty, line 138: The result should be generated per agenda.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+    CreateVotePaper(TxCreateVotePaper),
+    Vote(TxVote),
+}
+
+impl Action for TxAction {}
+pub type OwnTransaction = crate::common::SignedTransaction<TxAction>;
+
+impl Context {
+    fn excute_tx(&mut self, transaction: &Transaction) -> Result<H256, ExecuteError> {
+        if transaction.tx_type() != "Vote" {
```
majecty, line 225: Let's make a constant for the "Vote".

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                general_meeting_id,
+                agenda_number,
+                number_of_shares,
+                voter_name: voter_name,
+                voter_public_key,
+            }) => {
+                let meeting: GeneralMeeting = self.meeting_mut().get_meeting(&general_meeting_id).unwrap();
+                if meeting.id != general_meeting_id {
+                    return Err(ExecuteError::InvalidMeeting)
+                }
```
majecty, line 243: This comparison is strange. How the meeting has a differnt id when we query the meeting with the id.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+
+        match tx.tx.action {
+            TxAction::CreateVotePaper(TxCreateVotePaper {
+                general_meeting_id,
+                agenda_number,
+                number_of_shares,
+                voter_name: voter_name,
+                voter_public_key,
+            }) => {
+                let meeting: GeneralMeeting = self.meeting_mut().get_meeting(&general_meeting_id).unwrap();
```
majecty, line 240: Please do not use `unwrap` here.
It will make Voting module crash when the CreateVotePaper transaction has an invalid general_meeting_id.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                if meeting.id != general_meeting_id {
+                    return Err(ExecuteError::InvalidMeeting)
+                }
+                if meeting.tallying_time.time < self.block_header.as_ref().unwrap().timestamp() {
+                    return Err(ExecuteError::InvalidMeetingTime)
```
majecty, line 245: The error name looks not good.
When I read the error name I thought that the time in the general meeting is invalid.

Actually, the `tallying_time` in the `GeneralMeeting` is valid. What is incorrect is that the vote paper is created after tallying. Please change the error name.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                }
+                if meeting.number_of_agendas < agenda_number {
+                    return Err(ExecuteError::InvalidNumberofAgendas)
+                }
```
majecty, line 249: Does the agenda number start from 0? or 1?

If it starts from 0, this condition is wrong. The condition should be `meeting.number_of_agendas <= agenda_number`.
If it starts from 1, this condtion is wrong. The condition should be `0 < agenda_number && agenda_number <= meeting.number_of_agendas`.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                if meeting.voting_end_time.time > self.block_header.as_ref().unwrap().timestamp() {
+                    return Err(ExecuteError::InvalidVotingTime)
```
majecty, line 251: Please the name of the error.
`VotingTime` is correct. The problem is that `CreateVotePaper` transaction is included in a block too late.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+            TxAction::Vote(TxVote {
+                vote_paper_id,
+                choice,
+                voter_signature,
+            }) => {
+                if choice != favor || choice != against || choice != absention {
+                    return Err(ExecuteError::InvalidChoice)
+                }
+                let mut vote_paper: VotePaper =
+                    serde_cbor::from_slice(&self.storage().get(vote_paper_id.0.as_ref()).unwrap()).unwrap();
```
majecty, line 273: What happens if there is no VotePaper?

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                let corresponding_paper_id = vote_paper.vote_paper_id;
+                let vote = Vote::new(vote_paper_id, choice, voter_signature);
+                let vote_id = vote.clone().vote_id;
+                vote_paper.set_used_in(vote_id.clone());
```
majecty, line 277: Please check the vote_paper.used_in is already set or not.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                    voter_public_key,
+                );
+                self.storage_mut().set(vote_paper.vote_paper_id.0.as_ref(), serde_cbor::to_vec(&vote_paper).unwrap());
+                Ok(vote_paper.vote_paper_id)
+            }
+
+            TxAction::Vote(TxVote {
+                vote_paper_id,
+                choice,
+                voter_signature,
```
majecty, line 267: Please check the voter_signature is generated from the voter private key.
Since we have voter public key, we can verify it.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                agenda_number,
+                voter_name,
+                number_of_shares,
+                voter_public_key,
+            }) => {
+                let meeting: GeneralMeeting =
+                    serde_cbor::from_slice(&self.storage().get(general_meeting_id.as_ref()).unwrap()).unwrap();
+                if meeting.id != general_meeting_id {
+                    return Err(ExecuteError::InvalidMeeting).map_err(|_| todo_fixthis)
+                }
```
majecty, line 337: When the returned meeting.id != general_meeting_id?

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+        match tx.tx.action {
+            TxAction::CreateVotePaper(TxCreateVotePaper {
+                general_meeting_id,
+                agenda_number,
+                voter_name,
+                number_of_shares,
+                voter_public_key,
+            }) => {
+                let meeting: GeneralMeeting =
+                    serde_cbor::from_slice(&self.storage().get(general_meeting_id.as_ref()).unwrap()).unwrap();
```
majecty, line 334: Please handle the case when the GeneralMeeting does not exist.

votingsystem/src/voting.rs
```diff
@@ -0,0 +1,419 @@
...
+                if meeting.id != general_meeting_id {
+                    return Err(ExecuteError::InvalidMeeting).map_err(|_| todo_fixthis)
+                }
+                if meeting.number_of_agendas < agenda_number {
```
majecty, line 338: Does the agenda number start from 0? or 1?

If it starts from 0, this condition is wrong. The condition should be `meeting.number_of_agendas <= agenda_number`.
If it starts from 1, this condtion is wrong. The condition should be `0 < agenda_number && agenda_number <= meeting.number_of_agendas`.

votingsystem/src/voting.rs
MSNTCS: @majecty set_used_in function checks this.

majecty: You should check the returned result of the `set_used_in` function.
If you did not check the returned value, actually you didn't check the property.

votingsystem/src/voting.rs
MSNTCS: @majecty it is started from 0. if the given number of agendas is equal to the actual number of agendas; there should not be any error the error happens when the given agenda is greater than the actual one.

votingsystem/src/voting.rs
MSNTCS: @MSNTCS The verification happens when `execute_tx` is called.

voting/src/general_meeting.rs
```diff
@@ -0,0 +1,448 @@
...
+                    serde_cbor::from_slice(&transaction.body()).map_err(|_| ExecuteError::InvalidFormat)?;
+
+                let meeting_id = tx.action.meeting_id;
+                let key = crate::generate_voting_box_key(&meeting_id);
+                let vote_box: VoteBox = serde_cbor::from_slice(&self.storage().get(key.as_slice()).unwrap()).unwrap();
+                let mut vote_result = VoteResult::new();
+
+                let mut meeting: GeneralMeeting =
+                    serde_cbor::from_slice(&self.storage().get(meeting_id.as_ref()).unwrap())
+                        .map_err(|_| ExecuteError::InvalidMetadata)?;
```
majecty, line 147: What is the the `Metadata`?

majecty, line 147: Could you split this line into multiple lines? Since there are multiple `unwrap`s, it is hard to read the role of unwraps.

I also changed the error handling. We should handle the case when the general meeting does not exist.
We don't need to handle the case when the serde deserialization fails. Since the state is only readable and writable by this module we can trust the deserialization.

```suggestion
                let mut meeting: GeneralMeeting = {
                    let bytes =
                        self.storage().get(meeting_id.as_ref()).ok_or_else(|| ExecuteError::GeneralMeetingNotFound)?;

                    serde_cbor::from_slice(&bytes)
                    .expect("General meeting is serialized by this code")
                };
```

voting/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "codechain-votingsystem"
+version = "0.1.0"
+authors = ["CodeChain Team <hi@codechain.io>"]
+edition = "2018"
+
+[dependencies]
+ctypes = { package = "codechain-types", path = "../types" }
+coordinator = { path = "../coordinator" }
+ccrypto = { package = "codechain-crypto", git = "https://github.com/CodeChain-io/rust-codechain-crypto.git", version = "0.2" }
```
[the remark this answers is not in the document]
MSNTCS: > If you don't know an overview of how dependencies work in general, I can explain it to you briefly.

I know it, thanks

voting/src/common.rs
```diff
@@ -0,0 +1,71 @@
...
+            Ok(())
+        } else {
+            Err(())
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct UserTransaction<T: Action> {
+    pub seq: u64,
```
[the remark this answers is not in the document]
MSNTCS: Now that I am thinking about it, we do not need it.

MSNTCS:
This PR is split into several small PRs. The list of PRs is as follows: 
 #459 ~ #464 and #466 ~ #472.

PR #454, opened 2020-08-03
Fix bug in app_desc validator

junha1:


dynaxis:
What a silly bug it is? I'll merge and later squash it into the previous commit. Thanks.

PR #457, opened 2020-08-09
Prepare for primitive update

sgkim126:
I'll upgrade primitives to support mold soon. The hashes defined in the new primitive don't have some methods that currently have.
This PR prepares the upgrade.


sgkim126:
@majecty I fixed format errors.

PR #458, opened 2020-08-11
Use codechain-crypto from v0.2.0 tag

majecty:
If we do not specify any tag, branch, or commit hash, Cargo uses the
master branch. So before this commit, if the master branch changes the
version to 0.3, this dependency does not work.


majecty:
We are not reviewing mold2 branch.

PR #462, opened 2020-08-12
Implement check_transaction to verify CreateGeneralMeeting transaction

MSNTCS:


voting/src/general_meeting.rs
```diff
@@ -35,6 +35,7 @@ enum ExecuteError {
     TallyingTimePassed,
     TallyingIsBeforeVotingEnd,
     VotingEndPassed,
+    InvalidAdmin,
```
majecty, line 38: Actually admin is not invalid. The transaction signer is not admin.
How about `NotAuthorized` or `NotAdmin`?

voting/src/general_meeting.rs
```diff
@@ -147,7 +148,24 @@ impl TxOwner for Context {
...
+        let todo_fixthis: coordinator::types::ErrorCode = 3;
+        match transaction.tx_type() {
+            CREATE_GENERAL_MEETING_TX_TYPE => {
+                let tx: CreateGeneralMeetingOwnTransaction =
+                    serde_cbor::from_slice(&transaction.body()).map_err(|_| todo_fixthis)?;
+                tx.verify().map_err(|_| todo_fixthis)?;
+
+                let admin_public_key = tx.signer_public;
+                let written_admin: Public =
+                    serde_cbor::from_slice(&self.storage().get(ADMIN_STATE_KEY.as_bytes()).unwrap()).unwrap();
```
majecty, line 160:
```suggestion
                let written_admin: Public = self.admin();
```

voting/src/general_meeting.rs
```diff
@@ -147,7 +148,24 @@ impl TxOwner for Context {
...
+
+                if written_admin != admin_public_key {
+                    return Err(ExecuteError::InvalidAdmin).map_err(|_| todo_fixthis)
+                }
```
majecty, line 164: I missed this in `excute_tx`. Could you add this check in `excute_tx`?

voting/src/general_meeting.rs
```diff
@@ -147,7 +148,24 @@ impl TxOwner for Context {
...
+                Ok(())
```
majecty, line 165: Please add this check:
```rust
                if tallying_time.time < voting_end_time.time {
                    return Err(ExecuteError::TallyingIsBeforeVotingEnd)
                }
```

MSNTCS: It is added.

PR #463, opened 2020-08-12
Implement a service to get an access to the general_meeting module state

MSNTCS:


voting/src/general_meeting.rs
```diff
@@ -186,6 +186,29 @@ impl TxOwner for Context {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub enum Error {
+    InvalidMeeting,
```
majecty, line 191: How about `MeetingNotFound` or `AbsentMeeting` or something?
The current name `InvalidMeeting` does not give the reader about what was wrong.

voting/src/general_meeting.rs
```diff
@@ -186,6 +186,29 @@ impl TxOwner for Context {
...
+
+impl GeneralMeetingManager for Context {
+    fn get_meeting(&mut self, meeting_id: &GeneralMeetingId) -> Result<GeneralMeeting, Error> {
+        if !self.storage().has(meeting_id.as_ref()) {
+            return Err(Error::InvalidMeeting)
+        }
+        let meeting: GeneralMeeting = {
+            let bytes = self.storage().get(meeting_id.as_ref()).ok_or_else(|| Error::InvalidMeeting)?;
+            serde_cbor::from_slice(&bytes).expect("General meeting is serialized by this code")
+        };
```
majecty, line 207: Since you checked the existence of the meeting using the `storage().has`, you don't need to check the error once more.
How about this:
```rust
             let bytes = self.storage().get(meeting_id.as_ref()).expect("We checked the existence in the above if statement");

```

voting/Cargo.toml
```diff
@@ -9,6 +9,7 @@ ccrypto = { package = "codechain-crypto", git = "https://github.com/CodeChain-io
 ckey = {package = "codechain-key", path = "../key"}
```
majecty, line 9: The commit message seems not correct.
`Implement a service for the general_meeting module ` says that you implemented a service for the general meeting module. However, the service is used by other module and it is not used in the general meeting module.

PR #466, opened 2020-08-13
Implement CreateVotePaper transaction

MSNTCS:


voting/src/voting.rs
```diff
@@ -46,6 +76,38 @@ impl Context {
...
+                tx.verify().map_err(|_| ExecuteError::InvalidSignature)?;
+
+                let meeting_id = tx.tx.action.general_meeting_id;
+                let name = tx.tx.action.voter_name;
+                let shares = tx.tx.action.number_of_shares;
+                let agendas = tx.tx.action.agenda_number;
+                let voter_pub_key = tx.tx.action.voter_public_key;
+
+                let meeting: GeneralMeeting =
+                    self.meeting_mut().get_meeting(&meeting_id).map_err(|_| ExecuteError::InvalidMeeting)?;
```
majecty, line 94: `get_meeting` returns error only when there is no meeting data using the meeting_id.
Let's change the error name more specific like "GeneralMeetingNotFound".

voting/src/voting.rs
```diff
@@ -46,6 +76,38 @@ impl Context {
...
+                if meeting.get_tallying_time().get_time() < self.block_header.as_ref().unwrap().timestamp() {
+                    return Err(ExecuteError::VoteAfterTallyingTime)
+                }
+                if 0 == agendas && agendas > meeting.get_agendas() {
+                    return Err(ExecuteError::InvalidNumberofAgendas)
```
majecty, line 99: When I read the `InvalidNumberofAgendas` error name, I understood that the number of agendas in a general meeting is wrong. However, it isn't.
agenda_number in the TxCreateVotePaper is wrong. How about `InvalidAgendaNumber`?

voting/src/voting.rs
```diff
@@ -36,6 +38,14 @@ impl Context {
     fn storage_mut(&mut self) -> &mut dyn SubStorageAccess {
         self.storage.as_mut().unwrap().as_mut()
     }
+
+    fn _meeting(&self) -> &dyn GeneralMeetingManager {
```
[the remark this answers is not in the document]
MSNTCS: Because I am not using it for now. Is it better to remove it?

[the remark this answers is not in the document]
majecty: Let's add it when it is used.

voting/src/voting.rs
```diff
@@ -39,13 +69,41 @@ impl Context {
         self.storage.as_mut().unwrap().as_mut()
     }
 
-    fn _meeting(&self) -> &dyn GeneralMeetingManager {
-        self.general_meeting.as_ref().unwrap().as_ref()
-    }
-
```
majecty: You should remove this in the previous commit.

voting/src/voting.rs
```diff
@@ -14,16 +14,46 @@
...
+    InvalidSignature,
+    InvalidFormat,
+    GeneralMeetingNotFound,
+    VoteAfterEndTime,
+    VoteAfterTallyingTime,
+    InvalidAgendaNumber,
+    UsedVotePaper,
+}
+
+const CREATE_VOTE_PAPER_TX_TYPE: &str = "Vote_Paper";
```
majecty, line 43: Let's use the constant value consistent.
```suggestion
const CREATE_VOTE_PAPER_TX_TYPE: &str = "Create_Vote_Paper";
```

PR #467, opened 2020-08-13
Implement sorting module

junha1:


majecty:
I asked the goal of this sorting module to Junha.

Sorting module in this PR is assuming that all transactions are signed by public key of account.
Some application does not meet this requirement.

timestamp/src/sorting.rs
```diff
@@ -0,0 +1,156 @@
...
+                    } else {
+                        accounts.insert(public, vec![(seq, i)]);
+                    }
+                } else {
+                    invalid.push(i);
+                }
+            } else {
+                invalid.push(i);
+            }
+        }
```
majecty, line 67: Let's split this loop into two parts.
* Getting (public, account)
* Using (public, account)

For example:
```rust
 impl Context {
     fn account_and_seq_from_tx(&self, tx: &TransactionWithMetadata) -> Option<(Public, u64)> {
         let get_account_and_seq: &Box<dyn GetAccountAndSeq> = match self.get_account_and_seqs.get(tx.tx.tx_type()) {
             Some(get_account_and_seq: &Box<dyn GetAccountAndSeq>) => get_account_and_seq,
             None => return None,
         };

         match get_account_and_seq.get_account_and_seq(&tx.tx) {
             Ok((public: Public, seq: u64)) => return Some((public, seq)),
             _ => return None,
         }
     }
 }


 impl TxSorter for Context {
     // TODO: Consider origin
     fn sort_txs(&self, txs: &[TransactionWithMetadata]) -> SortedTxs {
         // TODO: Avoid Public hashmap
         let mut accounts: HashMap<Public, Vec<(u64, usize)>> = HashMap::new();
         let mut invalid: Vec<usize> = Vec::new();

          for (i: usize, tx: &TransactionWithMetadata) in txs.iter().enumerate() {
             if let Some((public: Public, seq: u64)) = self.account_and_seq_from_tx(tx) {
                 if let Some(list: &mut Vec<(u64, usize)>) = accounts.get_mut(&public) {
                     list.push((seq, i));
                 } else {
                     accounts.insert(k: public, vec![(seq, i)]);
                 }
             } else {
                 invalid.push(i);
             }
         }


```

timestamp/src/sorting.rs
```diff
@@ -0,0 +1,156 @@
...
+        let mut sorted: Vec<usize> = Vec::new();
+
+        for (account, list) in accounts.iter_mut() {
+            list.sort();
+            let mut real_seq = if let Ok(x) = self.account_manager().get_sequence(account, true) {
+                x
+            } else {
+                let x: Vec<usize> = list.iter().map(|x| x.1).collect();
+                invalid.extend_from_slice(&x);
+                continue
```
majecty, line 78: It's hard to know what is happening.

How about this?
```suggestion
                let tx_indexes: Vec<usize> = list.iter().map(|(seq, tx_index)| tx_index).collect();
                invalid.extend_from_slice(&tx_indexes);
                continue
```

timestamp/src/sorting.rs
```diff
@@ -0,0 +1,156 @@
...
+use foundry_module_rt::UserModule;
+use parking_lot::RwLock;
+use remote_trait_object::raw_exchange::{import_service_from_handle, HandleToExchange, Skeleton};
+use remote_trait_object::{service, Context as RtoContext, Service};
+use std::collections::HashMap;
+use std::sync::Arc;
+
+#[service]
+pub trait GetAccountAndSeq: Service {
+    fn get_account_and_seq(&self, tx: &Transaction) -> Result<(Public, u64), ()>;
```
majecty, line 30: How about defining a type for the sequence?

timestamp/src/sorting.rs
```diff
@@ -0,0 +1,156 @@
...
+            } else {
+                invalid.push(i);
+            }
+        }
+
+        let mut sorted: Vec<usize> = Vec::new();
+
+        for (account, list) in accounts.iter_mut() {
+            list.sort();
+            let mut real_seq = if let Ok(x) = self.account_manager().get_sequence(account, true) {
```
majecty, line 73: How about `seq_in_state` instead of `real`?
The sequences used in transactions are not fake or unreal.

```suggestion
            let mut seq_in_state = if let Ok(x) = self.account_manager().get_sequence(account, true) {
```

timestamp/src/sorting.rs
```diff
@@ -0,0 +1,156 @@
...
+                x
+            } else {
+                let x: Vec<usize> = list.iter().map(|x| x.1).collect();
+                invalid.extend_from_slice(&x);
+                continue
+            };
+            if list[0].0 != real_seq {
```
majecty, line 80: I'm to sad to see `list[0].0`.
I don't have a good alternative.

timestamp/src/sorting.rs
```diff
@@ -0,0 +1,156 @@
...
+                let x: Vec<usize> = list.iter().map(|x| x.1).collect();
+                invalid.extend_from_slice(&x);
```
majecty, line 82: How about giving a good name instead of x?
It's hard to follow the code if we use x.

```suggestion
                let tx_indexes: Vec<usize> = list.iter().map(|(seq, tx_index)| tx_index).collect();
                invalid.extend_from_slice(&tx_indexes);
```

PR #468, opened 2020-08-13
Implement check_transaction to verify CreateVotePaper transaction

MSNTCS:


voting/src/voting.rs
```diff
@@ -139,7 +139,24 @@ impl TxOwner for Context {
...
     fn check_transaction(&self, transaction: &Transaction) -> Result<(), coordinator::types::ErrorCode> {
-        unimplemented!();
+        let todo_fixthis: coordinator::types::ErrorCode = 3;
+        match transaction.tx_type() {
+            CREATE_VOTE_PAPER_TX_TYPE => {
+                let tx: CreateVotePaperTransaction =
+                    serde_cbor::from_slice(&transaction.body()).map_err(|_| todo_fixthis)?;
+                tx.verify().map_err(|_| todo_fixthis)?;
+                let meeting_id = tx.tx.action.general_meeting_id;
+                let agendas = tx.tx.action.agenda_number;
```
majecty, line 149: Please check your code that it has the same problem as before.

https://github.com/CodeChain-io/foundry/pull/466#discussion_r469675774

> The name `agendas` looks not a good name.
> `agenda_number` is a number of agenda. If there are 5 agendas in a general meeting, there could be 5 vote papers for each agenda.
> `agendas` looks like there are multiple agendas in a voting paper. However, it is not true.

voting/src/voting.rs
```diff
@@ -139,7 +139,24 @@ impl TxOwner for Context {
...
+
+                let meeting: GeneralMeeting = serde_cbor::from_slice(&self.storage().get(meeting_id.as_ref()).unwrap())
+                    .map_err(|_| todo_fixthis)?;
+                if 0 == agendas && agendas > meeting.get_agendas() {
```
majecty, line 153: Sorry I missed this before.
`get_agendas` function name looks like it should return the list of agenda. However it is returning the number (or count) of agenda.
Please change the name to `get_number_ob_agenda`. Please make a new commit for it.

PR #469, opened 2020-08-13
Implement voting box to collect votes

MSNTCS:


voting/src/general_meeting.rs
```diff
@@ -356,3 +357,21 @@ impl GeneralMeeting {
         self.number_of_agendas
     }
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct VoteBox {
+    pub meeting_id: GeneralMeetingId,
+    pub votes: Vec<VoteId>,
+}
+impl VoteBox {
```
majecty, line 385:
```suggestion
}

impl VoteBox {
```

voting/src/general_meeting.rs
```diff
@@ -209,6 +210,17 @@ impl GeneralMeetingManager for Context {
         };
         Ok(meeting)
     }
+
+    fn get_vote_box(&mut self, key: &[u8]) -> Result<VoteBox, Error> {
+        if !self.storage().has(key) {
+            return Err(Error::MeetingNotFound)
```
majecty, line 219: Shouldn't we use `VoteBoxNotFound` instead of `MeetingNotFound`?

```suggestion
            return Err(Error::VoteBoxNotFound)
```

PR #479, opened 2020-08-20
Make module level do not clone the cache to read value

majecty:
WriteBack cache panic when it is cloned while it has a check point. This
commit makes ReadOnlyModuleLevelState borrow the cahce.

### DIFF
--- a/state/src/impls/top_level.rs
+++ b/state/src/impls/top_level.rs
@@ -103,7 +103,7 @@ impl TopStateView for TopLevelState {
         match self.module_root(storage_id)? {
             // FIXME: Find a way to use stored cache.
             Some(module_root) => {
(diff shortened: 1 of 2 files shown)


state/src/impls/top_level.rs
```diff
@@ -103,7 +103,7 @@ impl TopStateView for TopLevelState {
         match self.module_root(storage_id)? {
             // FIXME: Find a way to use stored cache.
             Some(module_root) => {
-                let module_cache = self.module_caches.get(&storage_id).cloned().unwrap_or_default();
+                let module_cache = self.module_caches.get(&storage_id);
```
majecty, line 106: @junha1 @dynaxis  The `cloned` makes panic when a module reads data from storage.

junha1 (approved):

PR #481, opened 2020-08-20
Increasing max_body_size make the test succeed

majecty:


majecty:
https://github.com/CodeChain-io/foundry/pull/486 fixed the problem correctly.

PR #485, opened 2020-08-24
Implement a test for PublishResult transaction

MSNTCS:


voting/tests/common/scenarios.rs
```diff
@@ -317,3 +330,77 @@ pub fn test_vote(ctx: &RwLock<Context>) {
...
+    let publish_result = tx_publish_result(general_meeting_id);
+    let early_publish_transaction = ctx.write().tx_owners.get_mut("general_meeting").unwrap().execute_transaction(&publish_result);
+    assert!(early_publish_transaction.is_err());
+
+    let third_block_hash: BlockHash = BlockHash::default();
+    let third_author: Public = Public::default();
+    let third_set_validators: Vec<Public> = Vec::new();
+    let third_extera_data: Bytes = Bytes::default();
+
+    let header = Header::new(third_block_hash, 29 , 0, third_author, third_set_validators, third_extera_data.clone());
```
majecty, line 397: IMO, no need to use new variables since they are using the default value.
```suggestion
    let header = Header::new(block_hash, 29, 0, author, validators.clone(), extera_data.clone());
```

voting/tests/common/scenarios.rs
```diff
@@ -317,3 +330,77 @@ pub fn test_vote(ctx: &RwLock<Context>) {
...
+    let admin: Ed25519KeyPair = Random.generate().unwrap();
+    let admin_key = admin.public();
+
+    ctx.write()
+        .init_genesises
+        .get_mut("general_meeting")
+        .unwrap()
+        .init_genesis(&serde_cbor::to_vec(&admin_key).unwrap());
+
+    let create_meeting = tx_create_general_meeting(&admin.public(), &admin.private(), 2, 12, 18);
```
majecty, line 348: It's hard to know which is the timestamp.
How about declaring good names to the timestamps used in this test?

majecty (changes requested):

voting/tests/common/scenarios.rs
```diff
@@ -317,3 +330,76 @@ pub fn test_vote(ctx: &RwLock<Context>) {
...
+
+    let general_meeting_id: H256 = serde_cbor::from_slice(&creat_meeting_outcome.events.get(0).unwrap().value).unwrap();
+
+    // This transaction will return an error due to an early request.
+    let publish_result = tx_publish_result(general_meeting_id);
+    let early_publish_transaction =
+        ctx.write().tx_owners.get_mut("general_meeting").unwrap().execute_transaction(&publish_result);
+    assert!(early_publish_transaction.is_err());
+
+    let header = Header::new(block_hash, 29, 0, author, validators.clone(), extera_data.clone());
```
majecty, line 396: It's hard to know the meaning of 29 in this test. Let's change the code that to express that timestamp is bigger than tallying_time.
```suggestion
    let header = Header::new(block_hash, tallying_time + 10, 0, author, validators.clone(), extera_data.clone());
```

majecty (changes requested):

majecty (approved):

Issue #447, opened 2020-07-30
Failures in tests


junha1:
In bdbb207,

1. Travis CI
```
Dave should be a validator when deposit enough
CodeChain(117) exited with code 255, null
      1) "after each" hook for "Dave should be a validator when deposit enough"

4. Nominate new candidate
      with enough deposit, but without delegation
        ✓ should keep possible authors after a term change (13761ms)
        2) "after each" hook for "should keep possible authors after a term change"
```

2. My local machine
```
  1) Shutdown test
       Partial shutdown
         Alphas should be next validators after a complete shutdown:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/junha/Projects/foundry/test/src/e2e.dynval/2/dv.shutdown.test.ts)
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)

  2) Shutdown test
       "after each" hook for "Alphas should be next validators after a complete shutdown":
     Error: Timeout period is expired while waiting connect each other
      at PromiseExpect.checkFulfilled (src/helper/promise.ts:54:19)
      at Context.<anonymous> (src/e2e.dynval/2/dv.shutdown.test.ts:303:23)
      at processImmediate (internal/timers.js:456:21)

```
in my machine

majecty:
I've run tests in bdbb207cca04c3dde8dda411eada1e68d80ee787 and 8334a0c3e665b85740052f9a5ce22d4d8fb3a9b8 commit multiple times. All the tests succeed.

I've used the command below:
```sh
yarn mocha -r ts-node/register --timeout 10000 src/e2e.dynval/2/dv.shutdown.test.ts
yarn mocha -r ts-node/register --timeout 10000 src/e2e.dynval/1/dv.m-m\'.test.ts
```

It seems that failure occurs not frequently.

MSNTCS:
The same failure tests happened several times in my machine using @majecty's suggested commands. I will check the logic of related implementations.

MSNTCS:
> The same failure tests happened several times in my machine using @majecty's suggested commands. I will check the logic of related implementations. I ran tests in `bc771a4`. 

The failure seems to be because of the informer API. 
```
Thread 'main' panicked at 'Cannot listen TCP port: Os { code: 48, kind: AddrInUse, message: "Address already in use" }', network/src/p2p/handler.rs:153
```

[...]

PR #502, opened 2020-09-08
Shoutdown running nodes after running shoutdown tests

MSNTCS:
This PR is supposed to address this issue #447.


test/src/e2e.dynval/2/dv.shutdown.test.ts
```diff
@@ -297,6 +297,7 @@ describe("Shutdown test", function() {
                 );
             }
         });
+        await Promise.all(getValidators().nodes.map(node => node.clean()));
```
majecty, line 300: How about moving this to afterEach? In test, clean code normally done in afterEach.
Please read document of mocha.

majecty: Cleaning up is good. However this test is skipped because of line 223's `it.skip`. It will not change any behavior now. Is it your intention?

majecty (changes requested):

test/src/e2e.dynval/2/dv.shutdown.test.ts
MSNTCS: > How about moving this to afterEach? In test, clean code normally done in afterEach.
> Please read document of mocha.

@majecty The reason I called it at the end of `Total Shutdown` tests is that the current structure of the test does not allow me to call `clean` in `afterEach`. Since each of the nodes is created inside of each test, they cannot be called out side by `afterEach`.

majecty: @MSNTCS Is there any reason that each test should run the CodeChains?
Can't it be done in `beforeEach`?

MSNTCS: @majecty I think each test runs a validator who runs a CodeChain node.

majecty: Sorry, I didn't deliver my intention clearly.
How about moving "initializaing CodeChain processes" in the `beforeEach` function?
If it possible, we can "deinitialize CodeChain processes" in the `afterEach` function.
I think it will make the test simple and hard to mistake.

test/src/e2e.dynval/2/dv.shutdown.test.ts
```diff
@@ -22,6 +22,7 @@ import { validators } from "../../../tendermint.dynval/constants";
 import { PromiseExpect } from "../../helper/promise";
 import CodeChain from "../../helper/spawn";
 import { fullyConnect, setTermTestTimeout, withNodes } from "../setup";
+import { getValidators } from "../../stakeholder";
```
majecty: It seems not related to this PR. Could you check it?

majecty (changes requested):

majecty (dismissed):

majecty:
@MSNTCS I think Mergify should merge this PR.
Could you check the current status; why it is not works? It seems that it is related to Travis.

MSNTCS:
> @MSNTCS I think Mergify should merge this PR.
> Could you check the current status; why it is not works? It seems that it is related to Travis.

Let me re-run the failed test. If it passed correctly, I guess mergify will merge it.

majecty:
@MSNTCS It is not merged 😢

MSNTCS:
> @MSNTCS It is not merged 😢

I am trying to find what is the problem. :(

majecty (approved):

Issue #87, opened 2020-01-07
Integrate Mergify


majecty:
CodeChain-io/CodeChain project makes the Mergify bot automatically merge a PR if the PR is reviewed and all tests are passed. We need it in this repo.

majecty:
I enabled Mergify.

majecty:
It is not working.

MSNTCS:
@majecty Can you list what you have done for this?

majecty:
I don't remember well but, I think I enabled Mergify in GitHub config.
![image](https://user-images.githubusercontent.com/709900/92542406-9f91ee00-f283-11ea-8e86-337f3f1d9086.png)

MSNTCS:
> I don't remember well but, I think I enabled Mergify in GitHub config.
> ![image](https://user-images.githubusercontent.com/709900/92542406-9f91ee00-f283-11ea-8e86-337f3f1d9086.png)

It seems mergify works. look at this #236 and it is merged with mergify.
<img width="1000" alt="Screen Shot 2020-09-09 at 10 47 05 AM" src="https://user-images.githubusercontent.com/50092326/92545013-39f53000-f28a-11ea-8539-8506dddf4138.png">

MSNTCS:
@majecty I checked mergify documentation and everything seems alright and I think there is one working example. Should I close this issue?

majecty:
@MSNTCS OK.

PR #503, opened 2020-09-10
Fix mergify auto merge

MSNTCS:
This will fix issue #87 . 
Previously, the mergify worked, but since we are using new Travis CI, we need to change `status-success` condition.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -5,7 +5,7 @@ pull_request_rules:
       - "#review-requested=0"
       - "#changes-requested-reviews-by=0"
       - "status-success~=^Actions - "
-      - status-success=continuous-integration/travis-ci/pr
+      - status-success=Travis CI - Pull Request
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -5,7 +5,7 @@ pull_request_rules:
       - "#review-requested=0"
       - "#changes-requested-reviews-by=0"
       - "status-success~=^Actions - "
-      - status-success=continuous-integration/travis-ci/pr
+      - status-success=Travis CI - Pull Request
```
majecty, line 8: Can you write a simple history of why we changed this in the commit message?

Maybe these information can be included.
* It was worked before.
* Not working now.
* We are suspecting this point and trying it.

majecty (changes requested):

majecty (approved):

PR #506, opened 2020-09-15
Implement GraphQL webserver

junha1:


graphql/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+// TODO: replace with the real Client.
+pub struct Client {}
+
+impl Client {
+    fn allocate_session(&self, _height: Option<u64>) -> SessionId {
+        123
+    }
+}
+
+pub struct ServerData {
```
majecty, line 42: How about ServerContext?
`Data` is too general and does not deliver any meaning.
`Context` is aloso a bit general, but we can know it contains some conifuration or other useful objects that is used in GraphQL Server.

If the data is only used in Server construction, we may name it ServerConfig.

graphql/src/handler.rs
```diff
@@ -0,0 +1,23 @@
...
+/// various GraphQL resolver for thehost level(chain, mempool, net...)
+/// which will be implemented in this crate as well.
+pub fn handle_gql_query<T: async_graphql::ObjectType + Send + Sync + 'static>(
+    runtime: &tokio::runtime::Handle,
+    root: T,
+    query: &str,
+    variables: &str,
+) -> String {
+    let variables = if let Ok(s) = (|| -> Result<_, ()> {
+        Ok(async_graphql::Variables::parse_from_json(async_graphql::serde_json::from_str(variables).map_err(|_| ())?)
```
majecty, line 11: How about?
```suggestion
        let json_variables = async_graphql::serde_json::from_str(variables).map_err(|_| ())?;
        let variables = async_graphql::Variables::parse_from_json(json_variables).map_err(|_| ())?;
        Ok(variables)
```

I feel hard to read code if a line contains multiple error handlings.

graphql/tests/common/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+pub fn create_handler() -> Box<dyn HandleGraphQlRequest> {
+    Box::new(GraphQlRequestHandler::new())
+}
+
+#[test]
+fn query_directly() {
+    let handler = create_handler();
+    let result = handler.execute(123, &format!("{{ account(name: \"{}\") {{ balance }} }}", "John"), "{}");
```
majecty, line 95: Why are you using `format!`? If we use r#"…", it will be simple.

majecty (changes requested):

graphql/src/lib.rs
junha1: The name `Data` is to matched with `actix-web`'s `Data`.
`server_data: web::Data<Arc<ServerData>>,`

graphql/tests/integration_test.rs
```diff
@@ -0,0 +1,129 @@
...
+
+use actix_rt;
+use actix_web::client::Client;
+use coordinator::module::HandleGraphQlRequest;
+use fgql::ServerData;
+use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::sync::Arc;
+
+fn graphql_handlers() -> HashMap<String, (bool, Arc<dyn HandleGraphQlRequest>)> {
```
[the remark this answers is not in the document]
junha1: `bool` means whether the handler needs session or not.
I gave a name `session_needed` here https://github.com/CodeChain-io/foundry/pull/506/files#diff-72d1e1d1cdcc399f4b43c818ccb2726aR64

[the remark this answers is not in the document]
junha1: You mean making a new type for `bool`? Or the tuple?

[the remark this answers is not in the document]
majecty: What I mean was:
```
type SessionNeeded = bool
// or
struct GraphQLRequestHandler {
  session_needed: bool;
  handler: Arc<dyn HandleGraphQlRequest>
}
```

graphql/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+            let mut body = web::BytesMut::new();
+            while let Some(chunk) = payload.next().await {
+                let chunk = chunk?;
+                // limit max size of in-memory payload
+                if (body.len() + chunk.len()) > MAX_SIZE {
+                    return Err(ErrorBadRequest("overflow"))
+                }
+                body.extend_from_slice(&chunk);
+            }
+            Ok(serde_json::from_slice(body.as_ref()).unwrap())
```
majecty, line 112: I think this is using a low-level API in the library.
I'll find a high-level API.

junha1: FYI, I referred to this code https://github.com/actix/examples/blob/master/json/src/main.rs#L31

majecty (approved):

PR #508, opened 2020-09-16
Refactor actix handlers to use Extractor

majecty:


graphql/src/lib.rs
```diff
@@ -154,3 +103,42 @@ pub fn run_server(server_data: ServerData, addr: SocketAddr) -> Result<Server> {
...
 }
+
+struct SessionAndHandler {
+    pub session: u32,
+    pub handler: Arc<dyn HandleGraphQlRequest>,
+}
+
+impl FromRequest for SessionAndHandler {
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self, Self::Error>>>>;
```
majecty, line 114: I don't know well about this type `Pin<Box<dyn Future`. I found an example and followed.

majecty: https://stackoverflow.com/questions/63308246/how-to-use-async-code-in-actix-web-extractors

junha1: AFAIK returning associated Future object is workaround of async trait. http://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/

graphql/src/lib.rs
```diff
@@ -18,18 +18,18 @@ mod handler;
 
 use actix_web::{
```
junha1, line 19: Where does the term `Extractor` come from?

graphql/src/lib.rs
```diff
@@ -154,3 +103,42 @@ pub fn run_server(server_data: ServerData, addr: SocketAddr) -> Result<Server> {
     });
     Ok(server.bind(addr)?.run())
 }
+
+struct SessionAndHandler {
+    pub session: u32,
```
junha1, line 108: We have `SessionId`

graphql/src/lib.rs
majecty: From actix guide. They call `impl FromRequest` as extractor: https://actix.rs/docs/extractors/

junha1: Nice

graphql/src/lib.rs
majecty: Thanks!

graphql/src/lib.rs
majecty: Thanks, I changed the code to use it.

PR #519, opened 2020-10-05
Recover CI

junha1:


.github/workflows/build.yml
```diff
@@ -0,0 +1,35 @@
+on: [push, pull_request]
```
majecty, line 1: @junha1 Does the "Recover CI" commit have the same contents as before?
If it has the same contents as before, I'll not read the commit in detail.

majecty (changes requested):

.github/workflows/build.yml
junha1: It is same except `1.45.2` -> `1.46.0`

majecty (approved):

PR #524, opened 2020-10-06
Setup Foundry demo

junha1:


demo/tests/integration_test.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[actix_rt::test]
+async fn ping() {
+    let mut child = foundry_demo::run_node(5555);
+    sleep(Duration::from_secs(3));
+    let x = foundry_demo::request_query(5555, "ping", "aaaa", "aaaa").await;
+    assert_eq!(x, "Module not found: ping");
+    child.kill().unwrap();
+    child.wait().unwrap();
+}
```
majecty, line 39: Will you run 4 Foundry instances in tests?
I think creating a binary is better than running 4 nodes in the test.
What is your plan?

majecty (changes requested):

demo/tests/integration_test.rs
junha1: I just planned to call `foundry_demo::run_node()` multiple times. I don't understand what exactly `creating a binary` mean.

majecty: What I mean is that creating a main function and call four `run_node` in the main function.
 I thought the demo will work more than a few tens of minutes. Running a test for more than few minutes looks awkward.

junha1: Then how to deal with multiple tests?

majecty: What are your scenarios?

It seems we have misunderstood each other.

I thought the demo will be like this:
1. Run 4 nodes and make them connect.
2. Send transactions periodically to the network and watch them.
I thought we would run Foundry nodes for days or weeks.

We need to make a consensus about what is our goal.
How about asking details for Kwangyul?

junha1: I thought that we would run the demo for 2 or 3 mins. I'll ask him.

junha1: @majecty What is the conclusion considering KwangYul's comment?

majecty: Let's make a binary that runs 4 nodes.
It will be easy to app developers to run the network and easy to kill the netwok by killing the process.

junha1:
@majecty I updated. (demo -> integration test)

majecty (approved):

PR #526, opened 2020-10-07
Implement engine level GraphQL

junha1:


junha1:
@majecty 
Note:
- I omitted hash for block query, because I have to implement resolver for `H256`. (It is in timestamp module and we will have a common library for this later)
- Perhaps we can use `Option<SessionId>` instead of `NO_SESSION` but this will change the coordinator-module interface, so quite a big change. We can do later.

majecty (approved):

PR #528, opened 2020-10-08
Remove min cost and max transaction size from scheme file

majecty:
They are not used.


test/src/primitives/src/value/H256.ts
```diff
@@ -6,7 +6,7 @@ const RLP = require("rlp");
 export type H256Value = H256 | string;
 
 /**
- * Handles 256-bit data. Used to express block hash, parcel hash, transaction hash, merkle root, etc.
```
junha1, line 9: What is a parcel hash?

majecty: Parcel is a collection of transactions. They are not used anymore.

junha1: ok

junha1 (approved):

PR #531, opened 2020-10-08
Remove Params in Scheme files

majecty:


json/src/scheme/params.rs
```diff
@@ -1,80 +0,0 @@
-// Copyright 2018-2020 Kodebox, Inc.
```
majecty, line 1: I thought about whether remove this or not.

We may need this code to serialize Consensus?  params.

We don't need it now. This file is in scheme folder, We may rewrite similar code easily.
So I removed this.

junha1:
There is a fmt error

majecty:
@junha1 Thanks, I fixed them.

json/src/scheme/params.rs
junha1: ok.

junha1 (approved):

majecty:
I'll merge it. `cargo run test` succeeds in my local machine.

PR #532, opened 2020-10-08
Remove common params

majecty:
CommonParams has similar fields as ConsensusParams. There was no code that reads CommonParams.

After this PR, I'll rename ConsensusParams to ChainParams.


junha1 (approved):
Nice.

PR #534, opened 2020-10-08
Implement transaction submission via GraphQL

junha1:


integration-test/tests/integration_test.rs
```diff
@@ -19,16 +19,9 @@
 
 extern crate foundry_integration_test as test_common;
```
majecty, line 20: "Move get_latest_block() inside the crate" commit does not move get_latest_block.  it removed it.

majecty (changes requested):

integration-test/tests/integration_test.rs
junha1: Oh. My mistake. I'll fix.

majecty (approved):

majecty (approved):

PR #537, opened 2020-10-12
Add a new test send_hello_txes()

junha1:


coordinator/src/lib.rs
```diff
@@ -116,31 +116,32 @@ impl Coordinator {
...
+                .unwrap_or_else(|| {
+                    sessions.push(0);
+                    (sessions.len() - 1, 0)
+                });
 
-        sessions[index] |= 1 << bit;
-        let session_id = bit + (SESSION_BITS_PER_SLOT * index) as SessionId;
+            sessions[index] |= 1 << bit;
+            bit + (SESSION_BITS_PER_SLOT * index) as SessionId
+        };
```
majecty, line 138: @junha1 Could you explain how session is using bitset?

integration-test/tests/integration_test.rs
```diff
@@ -91,3 +91,39 @@ async fn send_hello_tx() {
     let value: Value = serde_json::from_str(&query_result).unwrap();
     assert_eq!(value["data"]["account"]["seq"], 1);
 }
+
+#[actix_rt::test]
+async fn send_hello_txes() {
```
majecty, line 96: How about making the test name to express the purpose?
This test is checking the sequence management code. Could you update the name of the test?

majecty (changes requested):

coordinator/src/lib.rs
junha1: @dynaxis authored this.

dynaxis: It's a bit more complex than absolutely required. I want to assign the smallest integer ID available and manage unoccupied IDs by mapping them to 0 bits (naturally 1 bits for occupied IDs). The code can be implemented using `bool` vector instead.

integration-test/tests/integration_test.rs
junha1: @majecty Would you recommend me a nice name?

majecty: How about `sequence_management`?
We may make tests like `sequence_management_1`, `sequence_management_2`, ...

junha1: Ok. I'll update

junha1: ok

timestamp/README.md
```diff
@@ -0,0 +1,50 @@
+# Timestamp Modules
```
majecty, line 1: Could you create a new commit for this README.md file?
It is contained in " Force single thread for GraphQL tokio runtime " commit.

majecty (changes requested):

timestamp/README.md
junha1: OMG. My mistake. I'll remove it

majecty (approved):

PR #545, opened 2020-10-14
Set scheme's state root before calling check_genesis_root

majecty:
Before this commit, check_genesis_root was checking whether the DB
contains 0x0 or not.

Before this commit, Foundry fail to read the DB generated before since
the DB does not have the entity of 0x0. After this commit, Foundry
succeed to read the DB to run.


junha1 (approved):

majecty:
Unit tests succeed in my machine.

PR #546, opened 2020-10-14
Call update_current_validator_set in the block.open

majecty:
It was called in the enact function which is used to verify a
block. However it was not called when generating a block.

update_next_block_state is called in the close function. For the
consistency, this commit make the `open` function call the
update_current_validator_set.


junha1 (approved):

majecty:
Unit tests succeed in my local machine.

PR #553, opened 2020-10-15
Move Genesis from scheme to app descriptor

majecty:


coordinator/src/app_desc/bytes.rs
```diff
@@ -0,0 +1,223 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
dynaxis, line 1: The only use of this `Bytes` thing in the context of app desc is just to deserialize `[u8]` from a long sequence of hexadecimal digits. I wrote a helper function to achieve the same in parsing app desc, and now I feel we require such a functionality here and there and are duplicating similar code. In this sense, we need to put it where appropriate (maybe in a separate crate) and reuse where relevant.

Also I'm not sure a new type here really adds any value here other than affecting serialization/deserialization.

I might need to search a bit more to find a crate that fits best. But https://docs.rs/serde_with/1.5.1/serde_with/hex/struct.Hex.html seems a good candidate.

coordinator/src/app_desc/genesis.rs
```diff
@@ -2,34 +2,34 @@
...
+use super::hash::H256;
+use super::seal::Seal;
 use ckey::PlatformAddress;
+use serde::Deserialize;
 
 /// Scheme genesis.
-#[derive(Debug, PartialEq, Deserialize)]
+#[derive(Debug, PartialEq, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct Genesis {
```
dynaxis, line 26: Should we allow to specify this many parameters to construct a genesis block's header? This question may be my ignorance, but I think we should minimize what chain creators need to specify. Please enlighten me on this.

coordinator/src/app_desc/hash.rs
```diff
@@ -0,0 +1,132 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
dynaxis, line 1: Is there any reason to copy and paste this code?

coordinator/src/lib.rs
```diff
@@ -270,7 +275,14 @@ impl Initializer for Coordinator {
 
         let (validator_set, params) = services.init_consensus.init_consensus(session_id);
 
-        self.max_body_size.set(params.max_body_size() as usize).expect("this must be the first assignment");
+        // FIXME: Should we cache max body size?
```
dynaxis, line 278: I think we can safely remove the caching part. What you stated your concern on it on discord seems the right way to go. I understand `max_body_size` should be a parameter that can be changed over time.

coordinator/src/app_desc/bytes.rs
majecty: Thanks. I just moved it from the scheme code. I'll add a commit that replace it using the serde_with crate.

coordinator/src/app_desc/genesis.rs
majecty: I agree with you. Most of the fields are not important.
As far as I know, when we want to create a new version of the test chain, we changes a small data in genesis header to get a new genesis hash. We can easily ignore all blocks from different genesis hash.

How about remaining only the `extraData` field?

coordinator/src/app_desc/hash.rs
majecty: I just copied it from the scheme code that I removed following commits.

coordinator/src/lib.rs
majecty: Thanks! I'll do it.

coordinator/src/app_desc/bytes.rs
dynaxis: Actually I also have such a parsing facility at `coordinator/src/app_desc.rs:235`. It should be replaced altogether. Once you are done, I'll do the similar.

coordinator/src/app_desc/genesis.rs
dynaxis: That's more minimal than what I expected. If `extraData` is enough, then I'm ok with it.

coordinator/src/app_desc/bytes.rs
dynaxis: @majecty @junha1 uses `hex` crate and it also serves the hex conversion quite well. Let's use it rather than `serde_with` if our main concern is to deal with hex representation of byte sequence.

majecty:
@dynaxis Can I create a separate PR to fix those issues?
1. Remove Bytes serialization and use `hex`.
2. Remove genesis configurations except extraBytes.
3. Remove max_body_size cache.

dynaxis:
You mean you want to create 3 PRs? I'm ok if you think it's better.

dynaxis:
> @dynaxis Can I create a separate PR to fix those issues?
> 1. Remove Bytes serialization and use `hex`.
> 2. Remove genesis configurations except extraBytes.
> 3. Remove max_body_size cache.

You mean you want to create 3 PRs? I'm ok if you think it's better.

PR #554, opened 2020-10-15
Use TOML in app descriptor

majecty:


coordinator/src/app_desc/params.rs
```diff
@@ -92,9 +92,7 @@ impl Value {
     fn merge_params(&mut self, merger: &Merger) -> anyhow::Result<()> {
         match self {
             Self::String(s) => {
-                let toml_string = format!("\"{}\"", s);
-                let merged = merger.merge(&toml_string)?;
-                *self = TOMLValueDeserializer::deserialize(&merged)?;
+                *self = Value::String(merger.merge(&s)?);
```
majecty, line 95: @dynaxis Could you check this code?
I'm not sure the intention of the merge.

If the goal is replacing text values, we don't need to deserialize the value.
Current tests only replacing text values like "==={{hello}}===" to "===value===".

Do you have a plan to use the "merger" for complex types?
For example, we may make an array like "[{{hello}}, 3, 4]" to "["world", 3, 4]".

dynaxis: @majesty my intention was to have a string containing `{{...}}` and let it interpreted as number or so after template expansion. Theoretically it's possible to use it for values of complex types. But I expect such use cases be rare.

Anyway, as I told you before, places where template can be specified are quite limited since template expansion occurs after fully parsing app desc. So even if an app desc author do some trick to create a complex value out of simple string, it either results in a parsing error or a value of wrong type different from what the consumer of the value expects.

majecty: I'll add some test cases that parses array or object.
I'm not sure current TOML conversion can keep it's behavior or not.

YAML can treat some text without qutes as string, but TOML can't. TOML string value must have `""`.

dynaxis: This is not what I considered while tracking transition to TOML. YAML is definitely flexible here. What I intended was to specify a template as a string value and to turn it into whatever type of value after the template is expanded. I'm not sure if it's possible with TOML.

dynaxis: My intention is:
1. To expand a value as a template if it's a string, regardless of whether it's actually a template or not.
    * Note that `{{...}}` can only be embedded in a string
1. To parse the resulting string from 1 into whatever type the string represents.

This was natural with YAML but not with TOML. The requirements here are:

1. Let strings be strings if they are actually just non-template strings.
1. Let expanded strings parsable as numbers and other types.
1. It's not strictly required to parse the expanded strings as complex types such as map and array.

dynaxis: The rules I suggest are:
1. String literals will be expanded as templates regardless of whether they contain {{..}} or not. And the results will be quoted and parsed as string literals.
2. A string starting with `@` will be a template and the result will be parsed unquoted after expansion. 
3. If it definitely required to put `@` at the beginning of a string, putting `@@` will yield a `@`.

dynaxis: By employing the above scheme for TOML, we can achive the following benefits:
1. Nothing gets in the way when writing no template TOML.
2. Templates can expand into inline table or array.

majecty:
@dynaxis Could you review this PR?

coordinator/src/app_desc.rs
```diff
@@ -410,27 +410,37 @@ mod tests {
...
+a = "awesome-module/a.a"
+
+[host.imports."\\namespace"]
+"b.b" = "asdfsdaf-asdf"
+
+[transactions]
+great-tx = "awesome-module"
+
+[param-defaults]
+num-threads = "10"
```
dynaxis, line 441: This should better be an integer? Actually no code uses the actual value of 10, BTW.

dynaxis (approved):
LGTM.
The comment above is quite minor. I think you can modify at your own will and merge this PR.
Good job. Thanks.

coordinator/src/app_desc.rs
majecty: @dynaxis The type of `param_defaults` is `Namespzed<String>`. It requires the config to be a `String` value.
I think in YAML, `num-threads: 10` can be parsed both number or string.

https://github.com/CodeChain-io/foundry/blob/fe8940a27cb5505f01b7d6bdf30b39e9b26d25b6/coordinator/src/app_desc.rs#L151

dynaxis: @majecty Maybe it is better typed as `Namespaced<Value>`. Anyway, let's proceed as it is, and deal with any possible problem when something gets in our way.

Issue #564, opened 2020-10-26
Organize hex byte string parsing code


majecty:
We have several functions that parse hex encoded string in app descriptor code.

In the current master branch: 2ab7f655e7c5fdbc3b8b5905ae9f01c0a7b24482

* deserialize_h256 function
https://github.com/CodeChain-io/foundry/blob/2ab7f655e7c5fdbc3b8b5905ae9f01c0a7b24482/coordinator/src/app_desc.rs#L249
* Bytes struct
https://github.com/CodeChain-io/foundry/blob/2ab7f655e7c5fdbc3b8b5905ae9f01c0a7b24482/coordinator/src/app_desc/bytes.rs#L29
* H256, H520
https://github.com/CodeChain-io/foundry/blob/2ab7f655e7c5fdbc3b8b5905ae9f01c0a7b24482/coordinator/src/app_desc/hash.rs#L93

PR #565, opened 2020-10-26
Make app descriptor use consistent byte deserialization

majecty:
Fix https://github.com/CodeChain-io/foundry/issues/564


coordinator/src/app_desc/genesis.rs
```diff
@@ -31,26 +31,44 @@ pub struct Genesis {
     /// Block timestamp, defaults to 0.
     pub timestamp: Option<u64>,
     /// Parent hash, defaults to 0.
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_option_h256")]
```
majecty, line 35: Omitting `     #[serde(deserialize_with = "deserialize_option_h256")] ` is too easy since `H256` already implements `Deserialize`. We can't use the `Deserialize` implementation since it assert using "0x" prefix.
Writing `struct H256Wrapper(H256)` for the deserialization code may be better.
@dynaxis Do you have any idea about it?

dynaxis: Overall, I stil feel the current code a bit verbose than necessary. Can we create a trait named like `TryFromU8Slice`, which takes a `[u8]` parsed from a hexadecimal digits and outputs a value of the target type (such as `H256`)? Once we implement them for all types to be parsed from hexadecimal digits, we can write a wrapper type like `Hex<T: TryFromU8Slice>`, and use it here and there.

majecty: I'll try it.

majecty:
@dynaxis Could you check this commit? https://github.com/CodeChain-io/foundry/pull/565/commits/f60c34bcf8890c4f378ef7ef50ddc0225de122a5
I've not finished this PR yet. I tried to create a `Hex` struct as you said in the comment.

PR #567, opened 2020-10-26
Read max_body_size from state

majecty:
Previously coordinator cached max_body_size when initializing the
genesis block. It was not a correct implementation. When we modify the
max_body_size in the state DB, the cached content and the DB content
will not be the same.

This commit makes the coordinator read max_body_size from the DB. Now,
we only have a source of truth.


dynaxis (approved):
LGTM

PR #569, opened 2020-10-28
Simplify config

majecty:
We choose structconf as a new config library. The config code we were
using was too verbose. We manually merge CLI args and config file
values one by one. We have two separate config lists; one for CLI args
and the other one for a config file.

'structconf' is not the best library. It does not print a good error
message. It can't read two config files; we were using two files. one
for default values, one for override values. There is no mature
library that meets our needs. Since structconf has minimum boilerplate
code we may move other libraries when they became mature.


dynaxis (approved):
LGTM.

PR #572, opened 2020-11-05
Receive CLI args for modules

majecty:
I've failed to create an example to override app desc values.
We may need a integration test environment to test it.
I manually tested it by changing timestamp/app_desc.toml and printing it in module constructor.

### DIFF
--- a/foundry/main.rs
+++ b/foundry/main.rs
@@ -20,8 +20,9 @@ extern crate log;
 extern crate codechain_logger as clogger;
 
 use app_dirs::AppInfo;
-use clap::SubCommand;
+use clap::{values_t, SubCommand};
 use config::Config;
+use std::{collections::BTreeMap, str::FromStr};
 use structconf::StructConf;
 
 pub use crate::run_node::run_node;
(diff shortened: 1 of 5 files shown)


foundry/main.rs
```diff
@@ -60,14 +62,46 @@ fn run() -> Result<(), String> {
         .about("Foundry client")
         .subcommand(SubCommand::with_name("commit-hash").about("Print the commit hash of the source tree"));
 
+    let app = app.arg(
+        clap::Arg::with_name("module_arguments")
+            .short("D")
+            .help("module arguments. For example  ''")
+            .takes_value(true)
+            .multiple(true),
+    );
```
majecty, line 71: @dynaxis I tried to add `-D` in Config struct, but failed.
`structconf` does not support multiple arguments

foundry/main.rs
```diff
@@ -60,16 +61,47 @@ fn run() -> Result<(), String> {
...
+    let mut ret = BTreeMap::new();
+
+    if args.is_none() {
+        return ret
+    }
+
+    let vals: Vec<&str> = args.unwrap().collect();
+    for keyvalue in vals {
+        let splitted: Vec<&str> = keyvalue.split('=').collect();
+        if splitted.len() != 2 {
```
dynaxis, line 94: It does no harm to permit `=` in a value. One might need to specify `=`s in the value.

foundry/main.rs
```diff
@@ -60,16 +61,47 @@ fn run() -> Result<(), String> {
...
+            panic!("Invalid module arguments. Please use -D key=value format");
```
dynaxis, line 95: Is it typical to panic on wrong argument?

dynaxis (changes requested):
A minor change seems preferable.
Regarding the use of panic, I'm just curious if it's a usual practice when using `clap`.

foundry/main.rs
dynaxis: https://github.com/clap-rs/clap/blob/master/examples/15_custom_validator.rs

This seems a more legitimate way to deal with a custom value format.

majecty: It uses only validation.
How about using `values_t!(...).unwrap_or_else(|e| e.exit())`?
`exit` is more user friendly since it does not print unnecessary call stacks.
https://github.com/clap-rs/clap/blob/v2-master/examples/12_typed_values.rs#L48

dynaxis: Ok. I thought that the validation failure should trigger printing of proper error messages.

dynaxis (dismissed):
LGTM

dynaxis (approved):
LGTM

PR #574, opened 2020-11-09
Splitting app descriptor

majecty:
There are many commits.

The first two PRs are small fix PRs. They are not directly related to the current PR.
1\. Use ./app-desc.toml as default value of app_desc_path
2. Use better error message to parse app descriptor

Pre-work
3. Extract app descriptor value types
4. Create LinkDesc struct and read it from file 
5. Extract base parameter override code 

Use link descriptor instead of app descriptor
6. Override params in the link descriptor
7. Validate link descriptor
8. Use link descriptor in weaver 

Clear work
9. Remove unused configs in app descriptor
10. Update config files


foundry/run_node.rs
```diff
@@ -234,8 +234,13 @@ pub fn run_node(
...
 
-    let mut app_desc = AppDesc::from_str(&fs::read_to_string(&config.app_desc_path).unwrap()).unwrap();
-    app_desc.merge_params(&module_arguments).expect("error in merge params");
+    let mut app_desc = {
+        let app_desc_path = &config.app_desc_path;
+        let app_desc_string = fs::read_to_string(app_desc_path)
+            .map_err(|err| format!("Foundry failed to read app desc file from {}: {}", app_desc_path, err))?;
+        AppDesc::from_str(&app_desc_string).map_err(|err| format!("Foundry failed to parse app descriptor: {}", err))?
+    };
+    app_desc.merge_params(&module_arguments).map_err(|err| format!("Foundry failed to merge params {}", err))?;
```
dynaxis, line 243: Maybe the message can be made more descriptive. `Foundry failed to merge params you supplied into the app descriptor.` should be better wording?

coordinator/src/lib.rs
```diff
@@ -17,6 +17,8 @@
 extern crate codechain_logger as clogger;
 extern crate log;
 
+#[macro_use]
+mod config_common;
```
dynaxis, line 21: `config_common` is a good name, but maybe `desc_common` is also a good name?

foundry/run_node.rs
```diff
@@ -241,6 +241,13 @@ pub fn run_node(
         AppDesc::from_str(&app_desc_string).map_err(|err| format!("Foundry failed to parse app descriptor: {}", err))?
     };
     app_desc.merge_params(&module_arguments).map_err(|err| format!("Foundry failed to merge params {}", err))?;
+    let _link_desc = {
+        let link_desc_path = &config.link_desc_path;
+        let link_desc_string = fs::read_to_string(link_desc_path)
+            .map_err(|err| format!("Foundry failed to read link desc file from {}: {}", link_desc_path, err))?;
```
dynaxis, line 247: `Foundry failed to read a link desc at {}: {}` ?

coordinator/src/lib.rs
```diff
@@ -94,11 +94,11 @@ pub struct Coordinator {
 const SESSION_BITS_PER_SLOT: usize = mem::size_of::<SessionSlot>() * 8;
 
 impl Coordinator {
-    pub fn from_app_desc(app_desc: &AppDesc) -> anyhow::Result<Coordinator> {
+    pub fn from_app_desc(app_desc: &AppDesc, link_desc: &LinkDesc) -> anyhow::Result<Coordinator> {
```
dynaxis, line 97: `from_descs`?

dynaxis (commented):
Please consider my suggestions.

Though the app desc has been splitted into two, the weaving part still has the two things intermixed. Let's complete this work and let's think a bit more on how to clearly separate the weaving part into two.

foundry/run_node.rs
majecty: I agree with you. Only saying "Merging params" does not give enough information.

coordinator/src/lib.rs
majecty: Good. The module contains common code of app descriptor and link descriptor.

foundry/run_node.rs
majecty: Thanks!

majecty:
@dynaxis The reason why we need an app descriptor in weaver is the way we handle transactions. We use special fields for transactions in the app descriptor. I don't know why do we use special fields for transactions.

dynaxis:
> @dynaxis The reason why we need an app descriptor in weaver is the way we handle transactions. We use special fields for transactions in the app descriptor. I don't know why do we use special fields for transactions.

Initially, I thought every export and import would be explicit but it was proven to be too verbose to descriptor authors. I'll think more about how to solve the dilemma.

dynaxis (approved):
LGTM

PR #575, opened 2020-11-12
Give default values to config arguments

majecty:
I tried to remove `unwrap` in Foundry initialization code.
`unwrap` makes programmers read code difficult.

### DIFF
--- a/coordinator/src/weaver.rs
+++ b/coordinator/src/weaver.rs
@@ -70,7 +70,7 @@ impl Weaver {
         self.modules.reserve(app_desc.modules.len());
 
         let host_module = link_desc.get("host").ok_or_else(|| anyhow!("can't find host module in app descriptor"))?;
-        self.process_host(host_module)?;
+        self.process_host(host_module);
         self.process_modules(app_desc, link_desc)?;
         self.tx_owners =
             app_desc.transactions.iter().map(|(tx_type, module)| (tx_type.clone(), (**module).clone())).collect();
@@ -84,7 +84,7 @@ impl Weaver {
         Ok((linkables, self.services.write().take().unwrap()))
     }
 
-    fn process_host(&mut self, link: &link_desc::ModuleSetup) -> anyhow::Result<()> {
+    fn process_host(&mut self, link: &link_desc::ModuleSetup) {
         let (exports, init_exports) = Self::process_exports(&link.exports);
         let imports = Self::process_imports(&link.imports);
         let imports = RefCell::new(imports);
@@ -115,8 +115,6 @@ impl Weaver {
             exports,
             imports,
         });
-
-        Ok(())
     }
 
     fn process_modules(&mut self, app_desc: &AppDesc, link_desc: &LinkDesc) -> anyhow::Result<()> {
--- a/coordinator/src/weaver.rs
+++ b/coordinator/src/weaver.rs
@@ -180,7 +180,7 @@ impl Weaver {
                 ..
             } in imports.iter()
             {
-                exports.push(*export_ids.get(from).unwrap());
+                exports.push(*export_ids.get(from).expect("We checked whether exporter exist in validate"));
             }
             export_port.export(&exports);
         }
--- a/foundry/config/mod.rs
+++ b/foundry/config/mod.rs
@@ -418,6 +418,12 @@ impl<T: FromStr> FromStr for CommaSeparated<T> {
     type Err = T::Err;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "" {
+            return Ok(CommaSeparated {
+                inner: vec![],
+            })
+        }
+
         let tokens = s.split(',');
         let mut ret: Vec<T> = Vec::new();
 
@@ -429,5 +435,29 @@ impl<T: FromStr> FromStr for CommaSeparated<T> {
         Ok(CommaSeparated {
             inner: ret,
         })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::CommaSeparated;
+
+    #[test]
+    fn comma_separated_empty() {
+        let empty: CommaSeparated<String> = "".parse().unwrap();
+        let expected: Vec<String> = vec![];
+        assert_eq!(empty.inner, expected);
+    }
+
+    #[test]
+    fn comma_separated_one() {
+        let one: CommaSeparated<String> = "one".parse().unwrap();
+        assert_eq!(one.inner, vec!["one".to_string()]);
+    }
+
+    #[test]
+    fn comma_separated_two() {
+        let onetwo: CommaSeparated<String> = "one,two".parse().unwrap();
+        assert_eq!(onetwo.inner, vec!["one".to_string(), "two".to_string()]);
     }
 }
--- a/foundry/config/mod.rs
+++ b/foundry/config/mod.rs
@@ -304,8 +304,13 @@ pub struct Config {
     #[conf(negated_arg, no_short, long = "no-snapshot", help = "Disable snapshots")]
     pub snapshot_enable: bool,
 
-    #[conf(no_short, long = "snapshot-path", help = "Specify the snapshot directory path.")]
-    pub snapshot_path: Option<String>,
+    #[conf(
+        no_short,
+        long = "snapshot-path",
+        help = "Specify the snapshot directory path.",
+        default = "\"snapshot\".to_string()"
+    )]
+    pub snapshot_path: String,
 
     // FIXME: We don't have a method to make snapshot_expiration as None.
     // If snapshot_expiration is None, it means that snapshots do not expire.
--- a/foundry/config/mod.rs
+++ b/foundry/config/mod.rs
@@ -63,8 +63,8 @@ pub struct Config {
     )]
     pub instance_id: Option<usize>,
 
-    #[conf(no_short, long = "base-path", help = "Specify the base directory path on which the "db" and "keys" directory will be created.")]
-    pub base_path: Option<String>,
+    #[conf(no_short, long = "base-path", help = "Specify the base directory path on which the "db" and "keys" directory will be created.", default = "\".\".to_string()")]
+    pub base_path: String,
 
     #[conf(no_short, long = "db-path", help = "Specify the database directory path.")]
     pub db_path: Option<String>,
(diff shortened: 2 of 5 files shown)


coordinator/src/weaver.rs
```diff
@@ -81,7 +81,7 @@ impl Weaver {
 
         let linkables = self.modules.into_iter().map(|(_, link_info)| link_info.linkable.into_inner()).collect();
 
-        Ok((linkables, self.services.write().take().unwrap()))
+        Ok((linkables, self.services.write().take().expect("Services is an option type to return it here")))
```
dynaxis, line 84: IMHO, use of `unwrap()` in production code (that is, outside of testing code) is a legitimate use of it. It means "I know what I'm doing. It won't be `Err` or `None`."

Besides the point, the sentence given as argument to `expect` is a bit weird to me. Shouldn't we say `services holds a value here"?

majecty: 1. I thought using `unwrap` is okay only when it is obvious. For example, it is checked in the code a few lines above.

When I read `unwrap` in code, I need to read how the value is created to ensure it will not panic.
I wanted to reduce the time to understand the `unwrap`.

I thought `weaver` is using `Option` in a non-obvious way. The `services` field is `Some` in the middle of `weave` function, and become `None` after the `weave` function.

How do you read code and what do you do when you read unknown `unwrap`?
I'm happy to discuss this with you. It will be great if we find a common consensus.

2. I agree with you in the point that the message of expect is bad. I think "services holds a value here" is not so good since it does not give information to programmers who read code. I'll think about the message when we agree on using `expect`.

dynaxis: Currently weaver's use of `Option` is a bit unconventional. But that context should better be conveyed by some additional comments. To me, `expect()` is considered imposing the possibility of actual panic. So readers need to grok the message given to `expect()` to know no panic will be issued. But in the case of `unwrap()`, it tells readers no panic will occur.

majecty: OK, I got your point. I'll let the code use `unwrap()`

majecty: @dynaxis I fixed it: https://github.com/CodeChain-io/foundry/compare/b7c5340be5139bb75f41bb0e411f1935486bf0ab..df61a2256691d5d3a360b8919b0bcb8d33415334

dynaxis: @majecty there is at least one more place correspnding to this discusson, and it's still using `expect()`.

majecty: OK, I'll review the use cases.

coordinator/src/weaver.rs
```diff
@@ -95,7 +95,8 @@ impl Weaver {
                 let mut buffer = Vec::<u8>::new();
                 let cbor =
                     &mut serde_cbor::Serializer::new(serde_cbor::ser::IoWrite::new(std::io::Cursor::new(&mut buffer)));
-                data.erased_serialize(&mut erased_serde::Serializer::erase(cbor)).unwrap();
+                data.erased_serialize(&mut erased_serde::Serializer::erase(cbor))
+                    .expect("Serializing vector to cbor format always succeed");
```
dynaxis, line 99: This is one of such uses we discussed.

coordinator/src/weaver.rs
```diff
@@ -182,7 +181,7 @@ impl Weaver {
                 ..
             } in imports.iter()
             {
-                exports.push(*export_ids.get(from).unwrap());
+                exports.push(*export_ids.get(from).expect("We checked whether exporter exist in validate"));
```
dynaxis, line 184: This too. Should be `unwrap()`.

majecty, line 184: I think this `expect` is worth. It exposes the assumption that code use.

foundry/config/mod.rs
```diff
@@ -281,62 +334,61 @@ pub struct Config {
...
+            reseal_min_period: Duration::from_millis(self.reseal_min_period),
         })
     }
 
     pub fn rpc_http_config(&self) -> RpcHttpConfig {
         // FIXME: Add interface, cors and hosts options.
         RpcHttpConfig {
-            interface: self.jsonrpc_interface.clone().unwrap(),
-            port: self.jsonrpc_port.unwrap(),
+            interface: self.jsonrpc_interface.clone().expect("We set a default value in the Config struct"),
```
dynaxis, line 365: This too? Should better be replaced with a comment and `unwrap()`?

foundry/config/mod.rs
```diff
@@ -281,62 +334,61 @@ pub struct Config {
...
         }
     }
 
     pub fn rpc_ws_config(&self) -> RpcWsConfig {
         // FIXME: Add hosts and origins options.
         RpcWsConfig {
-            interface: self.ws_interface.clone().unwrap(),
-            port: self.ws_port.unwrap(),
-            max_connections: self.ws_max_connections.unwrap(),
+            interface: self.ws_interface.clone().expect("We set a default value in the Config struct"),
```
dynaxis, line 381: Same as above.

foundry/config/mod.rs
```diff
@@ -281,62 +334,61 @@ pub struct Config {
...
+            max_connections: self.ws_max_connections,
         }
     }
 
     pub fn informer_config(&self) -> InformerConfig {
         InformerConfig {
-            interface: self.informer_interface.clone().unwrap(),
-            port: self.informer_port.unwrap(),
-            max_connections: self.informer_max_connections.unwrap(),
+            interface: self.informer_interface.clone().expect("We set a default value in the Config struct"),
```
dynaxis, line 389: This too.

foundry/config/mod.rs
```diff
@@ -371,26 +423,25 @@ impl Config {
...
-            self.bootstrap_addresses.as_ref().map(|addresses| addresses.inner.clone()).unwrap_or_default();
+        let bootstrap_addresses = self.bootstrap_addresses.inner.clone();
 
         let whitelist = make_ipaddr_list(self.whitelist_path.as_ref(), "white")?;
         let blacklist = make_ipaddr_list(self.blacklist_path.as_ref(), "black")?;
 
         Ok(NetworkConfig {
-            address: self.interface.clone().unwrap(),
-            port: self.port.unwrap(),
+            address: self.interface.clone().expect("We set a default value in the Config struct"),
```
dynaxis, line 432: This too.

majecty: I changed it to use `unwrap()`.

foundry/config/mod.rs
majecty: I changed it to use `unwrap()`.

foundry/config/mod.rs
majecty: I changed it to use `unwrap()`.

coordinator/src/weaver.rs
majecty: https://github.com/CodeChain-io/foundry/pull/575#discussion_r528539613
I don't agree with you. It can fail if a programmer give wrong arguments.
The `expect` will give information that we need validation code before using weave.

I thought that if we use `unwrap`, when the `unwrap` fails, programmer will think that `weaver` does not check the validity of arguments. They will change the weaver to check validity of arguments. However, they should fix validation code. This expect will give the information.

coordinator/src/weaver.rs
majecty: I changed it to use unwrap.

foundry/config/mod.rs
majecty: I changed it to use unwrap()

coordinator/src/weaver.rs
dynaxis: My point is "Let's use `unwrap()` and add comments instead."

But if we regard `expect()` as a `unwrap()` with a message for debugging, it's better to revert this PR to the original state you proposed, rather than letting `expect()` intact in this portion only.

https://stackoverflow.com/questions/39477684/should-i-avoid-unwrap-in-production-application/39478185#39478185

According to the discussion above in the page, there doesn't seem a clear conclusion on what nuiance each of `unwrap()` and `expect()` conveys. There is a person in the above link, who has a very similar opinion to mine but also another similar to yours.

I think `expect()` is horribly named if it is designed just as a `unwrap()` with a message. Since string literals go to the final binary, and we will anyway look into the code from the stacktrace printed by `unwrap()`, my preference is to add comments instead of `expect()`. Kinda I like assertions can be completely stripped off in production.

But in languages like Java, preconditions are frequently checked with usual method calls (such as `require(...)`), which can't be stripped out in production. If we regard calls to `expect()` in a similar way, then we can revert to your original uses of `expect()` and set a rule that `expect()` is preferred over `unwrap()`.

Please don't misunderstand me here. Having read some articles, I became to form my current perception on `unwrap()` and `expect()`. But discussing with you and searching for what others think, I'm starting to think differently.

So my conclusion is that if we leave this use of `expect()` here, we should prefer `expect()` unless the intention of using `unwrap()` is obvious seeing around it.

majecty: Thanks, I'll check your comment tomorrow.

dynaxis: I tried to think a bit more on the issue after writing the above comment.

1. `unwrap()` and `expect()` don't seem designed with different nuances
1. Many people use `unwrap()` only when it's 100% sure that the result is `Ok` or `Some`. Otherwise, `expect(..)` is commonly used.
    * That's in line with my perception, but I was not so clear about cases when `expect(..)` is used. I just thought `expect()` may be used for intended panics, which are triggered as a result of proper error handling.
    * It's quite reasonable not to bother with specifying a message for very obvious sitations, where `unwrap()` is more suitable than `expect()`.
    * So your discrimination of when to use or not use `unwrap()` is quite reasonable, in that `unwrap()` should be confined to cases where it's 100% sure that the result is ok or some only seeing a narrow context surrounding the `unwrap()`. In other cases, `expect(..)` signifies some invariants are set remotely.
1. Another separate concern regaring `unwrap()` and `expect(..)` is whether we should leave string literals to the final binary to make clear what invariant is violated (that is, when using `expect(..)`).
    * Panic caused by `expect(..)` is basically unexpected, and I think it's not mandatory to print messages. Knowing the location and examining comments there should be generally enough.
    * But it's in some sense beneficial to provide an immediate message, especially when additional contexts are supplied such as some additional values.

So in conclusion, let's leave `unwrap()` where really obvious, and use `expect(..)` when it's not immediately clear from the surrounding context (I mean less than dozen of lines).

dynaxis: So this use of `expect(..)` seems legitimate as you pointed out.

dynaxis:
If you have no objection on when to use `expect(..) or `unwrap()`, everything looks good to me. I learned a valuable lesson reviewing this PR. Thank you for your effort.

dynaxis (approved):
LGTM

coordinator/src/weaver.rs
majecty: Thanks. I didn't think about the final binary perspective.

PR #576, opened 2020-11-16
Improve integration test code

majecty:


integration-test/src/lib.rs
```diff
@@ -57,6 +65,24 @@ pub fn run_node(port: u16) -> FoundryNode {
     }
 }
 
+fn create_log_files(id: usize) -> (File, File) {
```
dynaxis, line 68: It always bugs me whether this kind of return types are ok. Though `create_log_files` is invoked only once, it returns two `File`s in a tuple and we can't tell which one is what by just looking at the function signature. Sometimes uses of tuples as return values are good, since the type of each item in them makes obvious what is returned at each item.

dynaxis (commented):
LGTM. But since I commented on a frequently manifesting pattern, could you share your thought on it?

integration-test/src/lib.rs
majecty: Great! I'll fix it now. I agree with you giving meaningful name to fields are almost always good.

majecty: @dynaxis I fixed it: https://github.com/CodeChain-io/foundry/compare/933bf422a969ef22c230d7c2d5e8044dc866bad9..9029e67248e9db6e45ec3eb654f920a6dd0da31a#diff-599de0e2ed4a8c0373496c4cec6979b39739627d47ac224766099fb7cff1877dR76

dynaxis (approved):
LGTM

PR #577, opened 2020-11-17
Split foundry lib and main

majecty:
Currently, we can use only timestamp modules.
By splitting main and lib we can use other modules easily.


dynaxis (approved):
LGTM

PR #583, opened 2020-11-25
Add new tests running multiple nodes

junha1:


integration-test/src/lib.rs
```diff
@@ -66,23 +74,40 @@ pub fn run_node(
...
+        .arg(format!("{}", foundry_args.mem_pool_size))
+        .arg("--engine-signer")
+        .arg(foundry_args.engine_signer)
+        .arg("--port")
+        .arg(format!("{}", foundry_args.port))
+        .arg("--bootstrap-addresses")
+        .arg(if foundry_args.bootstrap_addresses.is_empty() {
+            "".to_owned()
+        } else {
+            foundry_args.bootstrap_addresses.iter().fold("".to_owned(), |acc, x| format!("{},{}", acc, x))[1..]
```
majecty, line 105: I think `join` is better.

```
            foundry_args.bootstrap_addresses.join(",")
```

majecty (dismissed):

integration-test/src/lib.rs
junha1: Thanks. I updated

majecty (approved):

PR #587, opened 2020-11-26
Do not use 0x prefix for Public key

majecty:


types/src/validator_set.rs
```diff
@@ -19,10 +19,11 @@ use primitives::H256;
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
 use serde::{Deserialize, Serialize};
 use std::ops::{Deref, DerefMut};
+use foundry_hex::Hex;
 
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub struct CompactValidatorEntry {
-    pub public_key: Public,
+    pub public_key: Hex<Public>,
```
majecty, line 26: It looks awkward. I think internal type and serialize/deserialize type should be seperated.

majecty:
I don't have enough time to finish this PR.

PR #592, opened 2020-12-08
Provide option to use stdio

junha1:


integration-test/src/lib.rs
```diff
@@ -65,19 +65,22 @@ pub fn run_node(
         config_path,
     }: RunNodeArgs,
     foundry_args: FoundryArgs,
+    use_stdio: bool,
```
majecty, line 68: How about setting this using environment argument?
Setting `use_stdio` for each tests seems awkward.
Printing log for test is not characteristic of tests.
A programmer should be able to change the log config in runtime.

majecty (approved):

PR #597, opened 2020-12-16
Fix bug in transaction execution

junha1:


core/src/block.rs
```diff
@@ -188,14 +188,14 @@ impl OpenBlock {
     pub fn execute_transactions(
         &mut self,
         block_executor: &dyn BlockExecutor,
-        mut transactions: Vec<Transaction>,
+        transactions: Vec<Transaction>,
```
majecty, line 191: Could you add a description in the commit?
* The effect of the problem in the big picture.
* What was the actual cause of the problem.
* How did you fix it? Is it a temporary fix?

majecty (changes requested):

majecty (approved):

PR #598, opened 2020-12-17
Add new GraphQL field to query tx directly

junha1:


graphql-engine/src/lib.rs
```diff
@@ -48,6 +48,10 @@ impl QueryRoot {
     async fn event(&self, tx_hash: GqlH256) -> async_graphql::Result<Vec<Vec<u8>>> {
         Ok(self.client.events_by_tx_hash(&ctypes::TxHash::from(tx_hash.0)).into_iter().map(|x| x.value).collect())
     }
+
+    async fn transaction(&self, tx_hash: GqlH256) -> Option<u64> {
```
majecty, line 52: The name seems ambiguous.
I think `transaction` should return the transaction information, not the block number.

I'm not familiar with the GraphQL, but how about these things?

1.
```rust
query block(txHash: "xxxxxx") {
  blockNumber
}
```

... adding other examples.

majecty (changes requested):

majecty (approved):

graphql-engine/src/lib.rs
junha1: We concluded to discuss this later, though he's got a point.

PR #600, opened 2020-12-21
Replace u8 array to GqlBytes for GraphQL

junha1:


integration-test/src/lib.rs
```diff
@@ -165,7 +165,7 @@ pub async fn get_event(port: u16, tx_hash: primitives::H256) -> Vec<Vec<u8>> {
     let query_result = request_query(port, "engine", query, &variables.to_string()).await;
     let value: Value = serde_json::from_str(&query_result).unwrap();
     let list = value["data"]["event"].as_array().unwrap();
-    list.iter().map(|event| event.as_array().unwrap().iter().map(|x| x.as_i64().unwrap() as u8).collect()).collect()
+    list.iter().map(|event| hex::decode(event.as_str().unwrap()).unwrap()).collect()
```
majecty, line 168: It's hard to read the change. Could you explain it to me?

majecty (changes requested):

integration-test/src/lib.rs
junha1: The json result has been changed from
`"[75,4,32,45]"`
to
`"4b04202d"`

majecty (approved):

